PR #94, opened 2019-04-10
Release

selul:


selul:
@preda-bogdan the build is failing.

preda-bogdan:
@selul I've fixed the regex it was messing the tests, I think we can include unit testing in the precommit hooks, just a suggestion.

PR #132, opened 2019-08-02
Feature/gif video tag

selul:


inc/tag_replacer.php
```diff
@@ -62,22 +62,97 @@ public function init() {
...
+	 * @param string $content The content to process.
+	 *
+	 * @return bool
+	 */
+	public function img_to_video( $image_url, $image_tag, &$content ) {
+		if ( $this->settings->get( 'img_to_video' ) === 'disabled' ) {
+			return false;
+		}
+
+		$file = get_headers( $image_url, 1 );
```
selul: we should avoid checking the content type in this way, checking the extension only is enough.

inc/tag_replacer.php
```diff
@@ -62,22 +62,97 @@ public function init() {
...
+		$link_svg = apply_filters(
+			'optml_content_url',
+			$image_url,
+			array('width' => 'auto',
+				'height' => 'auto',
+				'format' => 'svg',
+			)
+		);
+
+		$video_tag = '<video autoplay muted loop playsinline poster="' . $link_svg . '">' .
```
selul: we should preserve classes and custom attributes on replace.

inc/url_replacer.php
```diff
@@ -169,16 +172,22 @@ public function build_image_url(
...
-				'apply_watermark' => apply_filters( 'optml_apply_watermark_for', true, $url ),
-			]
-		);
+		$arguments = [
+			'signed'          => $this->settings->use_lazyload() ? false : $this->is_allowed_site,
+			'apply_watermark' => apply_filters( 'optml_apply_watermark_for', true, $url ),
+
+		];
+
+		if ( isset( $args['format'] ) && ! empty( $args['format'] ) ) {
```
selul: it makes more sense to me to have the format in `$args` rather `$arguments`. I have imagined `$arguments` those more like global settings applied to all the images, even if the variable name is not that intuitive 🙉

inc/app_replacer.php
```diff
@@ -337,13 +337,19 @@ public function can_replace_url( $url ) {
...
 		$url_parts = parse_url( $url );
 
 		if ( ! isset( $url_parts['host'] ) ) {
 			return false;
 		}
+		// https://mlynllm40tdp.i.optimole.com/HNUQD3PR0fo/w:auto/h:auto/q:auto/
+		// http://c37866cd.ngrok.io/wp-content/uploads/2019/07/68747470733a2f2f6d65646961312e7465
+		// 6e6f722e636f6d2f696d616765732f65616337663764383533346630383433656264373037313031623865663766
+		// 642f74656e6f722e6769663f6974656d69643d3131363939363038.gif
+		// $this->set_properties();
```
selul: let's remove this comment

inc/app_replacer.php
```diff
@@ -337,13 +337,19 @@ public function can_replace_url( $url ) {
...
 		if ( false === ( isset( $this->possible_sources[ $url_parts['host'] ] ) || isset( $this->allowed_sources[ $url_parts['host'] ] ) ) ) {
-			return false;
+			return false;                     // here
```
selul: the comment seems redundant

selul:
we should add here a sample PHPUnit test and also a cypress scenario for a testcase when the origin will have the GIF feature deployed.

inc/tag_replacer.php
```diff
@@ -62,22 +62,106 @@ public function init() {
...
+		$video_tag = str_replace(
+			[
+				'src=',
+				'<img',
+				'/>',
+			],
+			[
+				'original-src=',
+				'<video autoplay muted loop playsinline poster="' . $link_png . '"',
+				'><source src="' . $link_mp4 . '" type="video/mp4"><source src="' . $link_webm . '" type="video/webm">sss</video',
```
selul: that `sss` seems redundant here

inc/tag_replacer.php
```diff
@@ -62,22 +62,106 @@ public function init() {
...
+	public function img_to_video( $image_url, $image_tag, &$content ) {
+		if ( $this->settings->get( 'img_to_video' ) === 'disabled' ) {
+			return false;
+		}
+
+		if ( $this->is_valid_mimetype_from_url( $image_url, [ 'gif' => false ] ) ) {
+			return false;
+		}
+
+		$link_webp = apply_filters(
```
selul: We don't seem to use this anywhere.

inc/image.php
```diff
@@ -107,13 +109,17 @@ public function get_url( $params = [] ) {
 		if ( isset( $this->resize->get()['type'] ) ) {
 			$path_parts[] = $this->resize->toString();
 		}
-
+		// format is not a must that's what we are changing
```
selul: not sure what means this comment here ?

GrigoreMihai: some old comment from when i was modifying code to remember what i do, it means nothing for others, i will delete it

inc/tag_replacer.php
GrigoreMihai: it was a link in the video tag, it looks like it got removed from there i think we should put it back there

inc/tag_replacer.php
```diff
@@ -62,22 +62,98 @@ public function init() {
...
 
-			$src = $tmp = $is_slashed ? stripslashes( $images['img_url'][ $index ] ) : $images['img_url'][ $index ];
+			$src = $tmp = $is_slashed ? stripslashes( $images['img_url'][ $index ] ) : $images['img_url'][ $index ]; // remove \
 
 			if ( strpos( $src, $this->upload_resource['content_path'] ) === 0 ) {
-				$src = $tmp = untrailingslashit( $this->upload_resource['content_host'] ) . $src;
+				$src = $tmp = untrailingslashit( $this->upload_resource['content_host'] ) . $src;   // remove last /
 
-				$new_src                     = $is_slashed ? addcslashes( $src, '/' ) : $src;
+				$new_src                     = $is_slashed ? addcslashes( $src, '/' ) : $src;    // make //
```
preda-bogdan: Can we remove this comments also?

optimole-wp.php
```diff
@@ -84,7 +84,8 @@ function optml() {
 	}
 
 	if ( ! defined( 'OPTML_DEBUG' ) ) {
-		define( 'OPTML_DEBUG', ( defined( 'TEST_GROUND' ) && TEST_GROUND ) ? true : false );
+		// define( 'OPTML_DEBUG', ( defined( 'TEST_GROUND' ) && TEST_GROUND ) ? true : false );
```
preda-bogdan: We should revert to how it was before so we should uncomment this line and remove the one bellow

tests/test-generic.php
```diff
@@ -25,6 +25,6 @@ function test_version_exists() {
 	 */
 	function test_debug_mode() {
 
-		$this->assertFalse( OPTML_DEBUG );
+		$this->assertTrue( OPTML_DEBUG );
```
preda-bogdan: Also this should be as before it should be assert false, the tests were failing from the previous comment where we have explicitly `define( 'OPTML_DEBUG', true );` if we revert to the way it was for production it should work as expected.

inc/tag_replacer.php
GrigoreMihai: yes, sure

preda-bogdan:
> @preda-bogdan that regex modification seems to solve the issue

@GrigoreMihai 
I would revert the changes and make a new branch from master, just for this changes so we can treat them separately, also this introduces conflicts with the base branch.

inc/tag_replacer.php
```diff
@@ -62,20 +62,96 @@ public function init() {
...
+
 		foreach ( $images[0] as $index => $tag ) {
 			$width     = $height = false;
 			$crop = null;
 			$image_tag = $images['img_tag'][ $index ];
 
 			$is_slashed = strpos( $images['img_url'][ $index ], '\/' ) !== false;
 
-			$src = $tmp = $is_slashed ? stripslashes( $images['img_url'][ $index ] ) : $images['img_url'][ $index ];
+			$src = $tmp = $is_slashed ? stripslashes( $images['img_url'][ $index ] ) : $images['img_url'][ $index ]; // remove \
```
preda-bogdan: Remove unnecessary comment

inc/tag_replacer.php
```diff
@@ -62,20 +62,96 @@ public function init() {
...
 
 			if ( strpos( $src, $this->upload_resource['content_path'] ) === 0 ) {
-				$src = $tmp = untrailingslashit( $this->upload_resource['content_host'] ) . $src;
+				$src = $tmp = untrailingslashit( $this->upload_resource['content_host'] ) . $src;   // remove last /
```
preda-bogdan: Remove unnecessary comment

PR #134, opened 2019-08-14
Feat/Lazyload_background

preda-bogdan:
chore: cherry pick changes from f1ad065f29f9b4e503c27de6f8467fb6c1f86077


inc/admin.php
```diff
@@ -78,6 +88,7 @@ public function inline_bootstrap_script() {
 		
 		</style>
 		<script type="application/javascript">
+					document.documentElement.className += "optimole_has_js";
```
selul: I think we need a space before the class, in case that the className already contains one, it will concat them rather than adding a new class.  
```suggestion
					document.documentElement.className += " optimole_has_js";
```

inc/admin.php
```diff
@@ -273,6 +286,26 @@ class="button button-primary"><?php _e( 'Connect to OptiMole', 'optimole-wp' );
...
+	protected function add_background_lazy_css() {
+
+		$watchers = $this->settings->get_watchers();
+		if ( ! isset( $watchers[ Optml_Settings::WATCHER_TYPE_LAZYLOAD ] ) ) {
+			return;
+		}
+
+		$css = '';
+		foreach ( $watchers[ Optml_Settings::WATCHER_TYPE_LAZYLOAD ] as $selector ) {
+			$css .= 'html.optimole_has_js ' . $selector . ':not(.optml-bg-lazyloaded) { background-image: none !important; } ';
```
selul: can we make sure to escape the selector, using `esc_attr`

inc/settings.php
```diff
@@ -10,6 +10,10 @@ class Optml_Settings {
 	const FILTER_FILENAME = 'filename';
 	const FILTER_TYPE_LAZYLOAD = 'lazyload';
 	const FILTER_TYPE_OPTIMIZE = 'optimize';
+
+	const WATCHER_TYPE_LAZYLOAD = 'lazyload';
+	const WATCH_CLASS = 'class';
```
selul: do we still need `WATCH_CLASS` and  `WATCH_ID`, i'm not able to find their usage

inc/admin.php
```diff
@@ -59,6 +59,16 @@ public function inline_bootstrap_script() {
 		$bgclasses       = Optml_Lazyload_Replacer::get_lazyload_bg_classes();
 		$watcher_classes = Optml_Lazyload_Replacer::get_watcher_lz_classes();
 
+		$watchers = $this->settings->get_watchers();
+		$lazyload_bg_selectors = [];
+		foreach ( $watchers[ Optml_Settings::WATCHER_TYPE_LAZYLOAD ] as $selector ) {
+			$lazyload_bg_selectors[] = $selector;
+		}
+		foreach ( $bgclasses as $key ) {
```
selul: I'm not sure what this is used for? This was used for a different mechanism, in the `backgroundReplaceClasses` prop of the library. Is it doing the same thing?

inc/url_replacer.php
```diff
@@ -116,7 +116,9 @@ public function build_image_url(
 
 		$is_slashed = strpos( $url, '\/' ) !== false;
 
-		$url = $is_slashed ? stripslashes( $url ) : $url;
+		// We do a little hack here, for json unicode chars we first replace them with html special chars,
+		// we then strip slashes to normalize the URL and last we convert html special chars back to get a clean URL
+		$url = $is_slashed ? html_entity_decode( stripslashes( preg_replace( '/\\\u([\da-fA-F]{4})/', '&#x\1;', $url ) ) ) : $url;
```
selul: we would need also to add a PHPUnit test to test this scenario

inc/admin.php
preda-bogdan: yes

PR #145, opened 2019-09-23
Compatibility/smart slider

preda-bogdan:


inc/lazyload_replacer.php
```diff
@@ -200,6 +200,8 @@ public function can_lazyload_for( $url, $tag = '' ) {
 		if ( false === Optml_Filters::should_do_image( $url, self::$filters[ Optml_Settings::FILTER_TYPE_LAZYLOAD ][ Optml_Settings::FILTER_FILENAME ] ) ) {
 			return false;
 		}
+		$url = strtok( $url, '?' );
+
```
selul: can you send a pr just with this in a separate PR, along with the test? rest of them seems not ready yet.

inc/manager.php
```diff
@@ -481,7 +481,7 @@ public function process_urls_from_content( $html ) {
 	 * @return array
 	 */
 	public function extract_image_urls_from_content( $content ) {
-		$regex = '/(?:http(?:s?):)(?:[\/\\\\|.|\w|\s|-])*\.(?:' . implode( '|', array_keys( Optml_Config::$extensions ) ) . ')(?:\?{1}[\w|=|&|\-|\.|:|;]*)?/';
+		$regex = '/((?:http(?:s?):)(?:[\/\\\\|.|\w|\s|-])*\.(?:' . implode( '|', array_keys( Optml_Config::$extensions ) ) . ')(?:\?{1}[\w|=|&|\-|\.|:|;]*)?)|((?:[\/\\\\|.|\w|\s|-])*\.(?:' . implode( '|', array_keys( Optml_Config::$extensions ) ) . ')(?:\?{1}[\w|=|&|\-|\.|:|;]*)?)/';
```
selul: this seems not right if we try to do target in regex both with and without schema urls, why we don't use just the regex without the schema to search. 

Also, how the re-written urls will look, they will include the schema of the original url?

GrigoreMihai: ok, i will test that 

yes, they will include the schema

inc/lazyload_replacer.php
preda-bogdan: Solved in #147

inc/manager.php
preda-bogdan: @GrigoreMihai You can check the regex here https://regex101.com/ and add the link with the tests to the conversation.

GrigoreMihai: https://regex101.com/r/LNLqzC/1

nextend:
I'm the developer of Smart Slider 3, if I can help you with this feature let me know.

selul:
@nextend from our findings smart slider uses schemaless URLs for images, making impossible for optimole to target them and replace the URLs. 
Which is the best way to fix this with Smart Slider? 

If there any filter which we can use to alter the final src of an image from slider with our URL format before you output it in the HTML?

nextend:
Schemaless is a setting in Smart Slider. It is enabled by default, but can be disabled.

I do not know too much about your plugin. Basically your goal is to replace 's src attribute with your own url. Can you work with background images as well?

Under the hood the following happens in Smart Slider when you display a slider on the frontend:
1.  Checks if the slider's HTML cached and the cache is valid
2.  If the cache invalid ->  the HTML code of the slider generated and displayed
3.  If the cache valid -> just display the HTML code

In step 2, Smart Slider could apply_filters for the image url's, but that would mean that the result is cached. So if something changes in your plugin or your plugin deactivated, the cache would be still valid with wrong urls.

In step 3, Smart Slider could apply_filters, but you would need to find the image urls with regexp, which I think you want to avoid as it is not simple task. This applies when you regexp the whole document without the filter. If you need Smart Slider could give a list of the images used in that specific slider (We have this info cached for Jetpack).


I saw in the commits that you are looking for `data-desktop="" data-tablet =""` and such attributes. Those are used for a `display:none` image which will be used as background images. So the image tag itself never loads those images. It is there for user who wants their slide backgrounds to show for search engines. Based on your comment in #148, we might need to add that specific exclude class only these hidden images to skip them and you would be able to find other images and replace them with your SVGs.

selul:
@nextend can we turn the schema-less setting off from Optimole, like hooking to a filter that changes the returned value of the setting,  when people chose the serve images from our side? 

Thanks !

PR #146, opened 2019-09-23
Fix/lightbox compatibility

preda-bogdan:


selul:
we need a test to back up this change

inc/manager.php
```diff
@@ -332,6 +332,11 @@ function ( $url ) {
 		);
 
 		foreach ( $urls as $origin => $replace ) {
+			if ( strstr( $origin, '?' ) !== false ) {
```
selul: for me, it seems we are not targeting the root cause here and only patching things at the end. 

The root cause is the fact that we alter the original URL in the `do_url_replacement` method, which calls `html_entity_decode`. 

What I would suggest instead of this, is to remove `html_entity_decode` . and apply that after the slashed check https://github.com/Codeinwp/optimole-wp/blob/44a74a8d1f751c6940f758cc90e3a9a97841a95d/inc/manager.php#L326 and before sending it to the replacement URL function. 

What do you think @GrigoreMihai @preda-bogdan ?

GrigoreMihai: we thought about modifying the `normalize_urls` when we started fixing this but we tried to avoid that because it is also used in `extract_urls_from_json` and removing `html_entity_decode` from `normalize_urls` will cause some tests to fail

preda-bogdan: @selul I've submitted a change where we don't use `html_entity_decode` witch was added as part of this revision 2bae741d03cc475d076c531df2512e301dd5fbab and just remove `&quot;` from origin URL it.

The issue is that before we decoded the original URL and then removed chars like `"` or `;` witch could have been left from parsing the json.

Now we just leave the url as is and just remove extra `&quot;` that broke the replacement after `html_entity_decode` was removed.

Is this a better option for tackling  this do you have some other tests in mind that we should be aware of, as of now everything passes and the query URLs from json get covered as well.

preda-bogdan:
@selul submitted the changes and added a test for URLs with encoded query arguments inside the URL

PR #178, opened 2019-11-12
Compatibility/slider revolution

GrigoreMihai:
The issue with Slider Revolution is fixed here, the fix also needs and includes the lazyload exclude by class feature


inc/tag_replacer.php
```diff
@@ -137,7 +152,11 @@ public function process_image_tags( $content, $images = array() ) {
 
 		$image_sizes = self::image_sizes();
 		$sizes2crop  = self::size_to_crop();
-
+		self::$filters[ Optml_Settings::FILTER_TYPE_LAZYLOAD ][ Optml_Settings::FILTER_CLASS ] = apply_filters(
```
selul: we don't need to initialize this, it's already done here https://github.com/Codeinwp/optimole-wp/blob/d21e4628128d0e5215bc3a47adb02f28b799af4e/inc/app_replacer.php#L271

The filter is not required here, we can use this one instead to exclude some image tags in various compatibilities https://github.com/Codeinwp/optimole-wp/blob/d21e4628128d0e5215bc3a47adb02f28b799af4e/inc/app_replacer.php#L127

To keep consistency, we can add the saved classes list to this filter `optml_possible_lazyload_flags` and remove the `contains_banned_lazyload_class` method.

GrigoreMihai: ok, I modified it now

inc/admin.php
```diff
@@ -611,13 +611,15 @@ private function get_dashboard_strings() {
...
 				'filter_ext'                        => __( 'Image extension', 'optimole-wp' ),
+				'filter_class'                      => __( 'Image class', 'optimole-wp' ),
 				'exclude_title_optimize'            => __( 'Don\'t optimize images if', 'optimole-wp' ),
 				'exclude_title_lazyload'            => __( 'Don\'t lazyload images if', 'optimole-wp' ),
 				'exclude_filename_desc'             => __( 'Image filename contains', 'optimole-wp' ),
 				'exclude_url_desc'                  => __( 'Page url contains', 'optimole-wp' ),
 				'exclude_ext_desc'                  => __( 'Image extension is', 'optimole-wp' ),
 				'watch_title_lazyload'              => __( 'Lazyload background images for selectors:', 'optimole-wp' ),
 				'watch_desc_lazyload'               => __( 'You can add each CSS selector on a new line or separated by comma(,).', 'optimole-wp' ),
+				'exclude_class_desc'                => __( 'Image containes class', 'optimole-wp' ),
```
selul: should be 
```suggestion
				'exclude_class_desc'                => __( 'Image tag contains class', 'optimole-wp' ),
```

PR #180, opened 2019-11-12
Compatibility/builders background lazyload

GrigoreMihai:
@preda-bogdan


inc/compatibilities/elementor_builder.php
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	/**
+	 * Register integration details.
+	 */
+	public function register() {
+		add_filter(
+			'optml_lazyload_bg_selectors',
+			function ( $all_watchers ) {
+				$all_watchers = array_merge( $all_watchers, ['.elementor-widget-container', '.elementor-background-overlay', '.div-with-background > .elementor-column-wrap'] );
```
selul: @GrigoreMihai `.div-with-background` i dont think it's a standard elementor class, and .elementor-background-overlay is already used here https://github.com/Codeinwp/optimole-wp/blob/master/inc/lazyload_replacer.php#L86

GrigoreMihai: ok

PR #181, opened 2019-11-12
Compatibility/sliders background lazyload

GrigoreMihai:
@preda-bogdan


inc/compatibilities/master_slider.php
```diff
@@ -0,0 +1,46 @@
...
+				return $all_watchers;
+			}
+		);
+		$settings = new Optml_Settings();
+
+		if ( $settings->get( 'lazyload' ) === 'enabled' ) {
+			add_filter(
+				'optml_lazyload_filename_excluded',
+				function ( $array ) {
+					$array['blank.gif'] = true;
```
selul: here I think it's better to exclude at all the optimization on this, not just the lazyload.

selul:
@GrigoreMihai can you have a look at the conflicts also.

inc/compatibilities/master_slider.php
GrigoreMihai: ok

PR #187, opened 2019-11-29
fix: edge15 needs poly version of js library

preda-bogdan:
References ans solves Codeinwp/optimole_js_lib#72


GrigoreMihai:
don't merge it yet, it seems to have a small mistake (doing poly version all time)

GrigoreMihai:
it's ok now

PR #189, opened 2019-12-04
fix: blur remaining outside of an image after it gets lazyloaded

GrigoreMihai:
@preda-bogdan


preda-bogdan:
Fixes Codeinwp/optimole_js_lib#55

Issue #168, opened 2019-10-24
replacement improvement coverage


selul:
We need to improve the replacer mechanism to cover non-schema urls and those without domains. 

we can brainstorm about possible approaches.

preda-bogdan:
I think this can be closed.

PR #190, opened 2019-12-05
fix: proposed solution for #168

preda-bogdan:
This should target and replace all schema-less URLs without duplicating or re-processing the base URL.

Also contains the `<noscript>` whitespace change and has a small fix for encoded `?` in URLs


inc/manager.php
```diff
@@ -495,7 +504,7 @@ public function process_urls_from_content( $html ) {
 	 * @return array
 	 */
 	public function extract_image_urls_from_content( $content ) {
-		$regex = '/(?:http(?:s?):)(?:[\/\\\\|.|\w|\s|-])*(?:[' . Optml_Config::$chars . '])*\.(?:' . implode( '|', array_keys( Optml_Config::$extensions ) ) . ')(?:\?{1}[\w|=|&|\-|\.|:|;]*)?/';
+		$regex = '/(?:http(?:s?):)?(?:[\/\\\\|.|\w|-])*(?:[' . Optml_Config::$chars . '])*\.(?:' . implode( '|', array_keys( Optml_Config::$extensions ) ) . ')(?:[\?|%3F]{1}[\w|=|&|\-|\.|:|;|%]*)?/';
```
preda-bogdan: Changed the group (http:|https:) to optional, added encoded form for `?` and allowed extra char `%` after extension

You can test the regex here:
https://regex101.com/r/Dd47Y1/3
and here:
https://regex101.com/r/Dd47Y1/4

inc/manager.php
```diff
@@ -336,17 +336,26 @@ private function do_url_replacement( $html, $extracted_urls ) {
...
+				}
+
 				return $is_slashed ? addcslashes( $new_url, '/' ) : $new_url;
 			},
 			$urls
 		);
 
 		foreach ( $urls as $origin => $replace ) {
-			$html = preg_replace( '/(?<!\/)' . preg_quote( $origin, '/' ) . '/m', $replace, $html );
+			$html = preg_replace( '/(?<![\/|:])' . preg_quote( $origin, '/' ) . '/m', $replace, $html );
```
preda-bogdan: Changed the regex to not match if preceded by `:` this allows to not duplicate replacements

preda-bogdan:
@selul added replacement for relative and schemaless URL's  + unit tests
Let me know if you see any issues

PR #194, opened 2019-12-10
Conflicts/divi

preda-bogdan:
Should add a notice for possible issues with DIVI


selul:
this is not required when the static css file generation is disabled, so we need to properly show the message based on the environment. 
moreover, the tab is added by the Divi theme builder, not the Divi plugin, so the conditional is too vague. 

Also this [doc](https://docs.optimole.com/article/1086-why-optimole-does-not-optimize-all-the-images-from-my-site) can be updated with this use case aswell.

inc/conflicts/divi.php
```diff
@@ -0,0 +1,52 @@
...
+	/**
+	 * Determine if conflict is applicable.
+	 *
+	 * @return bool
+	 * @since   2.2.6
+	 * @access  public
+	 */
+	public function is_conflict_valid() {
+
+		if ( ! ( strcmp( wp_get_theme(), 'Divi' ) === 0 ||
```
preda-bogdan: `wp_get_theme()` returns an Object type and you are using it to compare to a string.
Also take note that it might be a child theme of Divi

GrigoreMihai: I changed it, it should be ok now

inc/conflicts/divi.php
```diff
@@ -0,0 +1,62 @@
...
+	 * Determine if conflict is applicable.
+	 *
+	 * @return bool
+	 * @since   2.2.6
+	 * @access  public
+	 */
+	public function is_conflict_valid() {
+		$show_message = false;
+		$theme = wp_get_theme( 'Divi' );
+		if ( $theme->exists() ) {
```
preda-bogdan: You can merge this 2 nested if statements with `&&`

inc/conflicts/divi.php
```diff
@@ -0,0 +1,62 @@
...
+			if ( strcmp( $theme->get( 'Name' ), 'Divi' ) === 0 ) {
+				$show_message = true;
+			}
+		}
+		if ( is_plugin_active( 'divi-builder/divi-builder.php' ) ) {
+			$show_message = true;
+		}
+		if ( get_template_directory() !== get_stylesheet_directory() ) {
```
preda-bogdan: Also merge this 2 nested if statements with `&&`, in the future  if there is no other logic required try to avoid nested if's

inc/conflicts/divi.php
```diff
@@ -0,0 +1,62 @@
...
+	 * @access  public
+	 */
+	public function is_conflict_valid() {
+		$show_message = false;
+		$theme = wp_get_theme( 'Divi' );
+		if ( $theme->exists() ) {
+			if ( strcmp( $theme->get( 'Name' ), 'Divi' ) === 0 ) {
+				$show_message = true;
+			}
+		}
```
preda-bogdan: You can merge this 2 nested if statements with `&&`

inc/conflicts/divi.php
```diff
@@ -0,0 +1,62 @@
...
+		if ( is_plugin_active( 'divi-builder/divi-builder.php' ) ) {
+			$show_message = true;
+		}
+		if ( get_template_directory() !== get_stylesheet_directory() ) {
+			if ( strcmp( wp_get_theme( get_template() )->get( 'Author' ), 'Divi' ) === 0 ) {
+				$show_message = true;
+			}
+		}
```
preda-bogdan: Also merge this 2 nested if statements with `&&`, in the future if there is no other logic required try to avoid nested if's

preda-bogdan:
The fix references this issue: https://secure.helpscout.net/conversation/1033143949/210696?folderId=2353201

Issue #195, opened 2019-12-11
Change URL structure


preda-bogdan:
**To Do**
Change URL structure to be prefixed by a unique token.
The token should be a hash of 8bits containing the `userkey` and the `domain`

References: [Clear Cache Feature](https://github.com/orgs/Codeinwp/projects/25#card-30438873)

selul:
I'm suggesting here the next approach for the hash key.

Basically, the signature format would need to be `<domain_hash>-<url hash>`

where: 
- domain_hash - are the first 5 bytes from the sha1 of image domain + app secret ( not token )
- url_hash - are the first 8 bytes from the sha1 of the source URL + app secret ( not token ) 

I think that using userkey + domain is not enough because we need to be able to purge cache per domain or per URL, but right now we will open the URL only purge for everyone.

what do you think?

PR #198, opened 2019-12-13
chore: added cache token for URLs

preda-bogdan:
References #195


preda-bogdan:
@selul I've made the discussed changes if you have time to take a look.

Issue #100, opened 2019-05-11
pinterest share compatibility issue


selul:
We can check the compatibility with pinterest pin plugins, which might share the lazyload placeholder. 

one example is Sassy Social Share

GrigoreMihai:
** Save images from site option from the official pinterest site works (takes both the placeholders and the original images)

I will add here the tested plugins, tag them with **works** or **doesn't work** and then create compatibilities where needed:

- [ ] Sassy Social Share - **doesn't work**

PR #202, opened 2019-12-20
Compatibility/sassy social share

GrigoreMihai:
@preda-bogdan It should be ok now

### DIFF
--- a/inc/compatibilities/sassy_social_share.php
+++ b/inc/compatibilities/sassy_social_share.php
@@ -16,13 +16,27 @@ class Optml_sassy_social_share extends Optml_compatibility {
 	function should_load() {
 		include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
 
-		return is_plugin_active( 'sassy-social-share/sassy-social-share.php' );
+		if ( is_plugin_active( 'sassy-social-share/sassy-social-share.php' ) ) {
+			$ss_options = get_option( 'heateor_sss' );
+			$ss_bars = ['vertical_re_providers', 'horizontal_re_providers'];
+			foreach ( $ss_bars as $key => $bar ) {
+				foreach ( $ss_options[ $bar ] as $index => $value ) {
+					if ( isset( $value ) && is_string( $value ) ) {
+						if ( strpos( $value, 'pinterest' ) !== false ) {
+							return true;
+						}
+					}
+				}
+			}
+		}
+		return false;
 	}
 
 	/**
 	 * Register integration details.
 	 */
 	public function register() {
+
 		add_action(
 			'wp_head',
 			function () {
--- a/inc/compatibilities/sassy_social_share.php
+++ b/inc/compatibilities/sassy_social_share.php
@@ -19,20 +19,22 @@ class Optml_sassy_social_share extends Optml_compatibility {
 		}
 		$ss_options = get_option( 'heateor_sss' );
 		$ss_bars = ['vertical_re_providers', 'horizontal_re_providers'];
-		if ( ! empty( $ss_options ) && is_array( $ss_options ) ) {
-			foreach ( $ss_bars as $key => $bar ) {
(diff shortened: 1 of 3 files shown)


preda-bogdan:
references #100

inc/compatibilities/sassy_social_share.php
```diff
@@ -0,0 +1,63 @@
...
+class Optml_sassy_social_share extends Optml_compatibility {
+	/**
+	 * Should we load the integration logic.
+	 *
+	 * @return bool Should we load.
+	 */
+	function should_load() {
+		include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+
+		if ( is_plugin_active( 'sassy-social-share/sassy-social-share.php' ) ) {
```
selul: let's return early here 
```suggestion
		if ( ! is_plugin_active( 'sassy-social-share/sassy-social-share.php' ) ) {
		    return false;
		}
```

inc/compatibilities/sassy_social_share.php
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	public function register() {
+		add_action(
+			'wp_head',
+			function () {
+				$output = '
+		<script type="application/javascript">
+			(function(w, d){ 
+					w.addEventListener("load", function(){
+						let p=d.querySelectorAll( ".heateorSssSharing.heateorSssPinterestBackground" );
```
selul: We can use wp_add_inline_script.

inc/compatibilities/sassy_social_share.php
GrigoreMihai: ok

inc/compatibilities/sassy_social_share.php
GrigoreMihai: using this will require to add an empty script above what we add

selul: no, we can use false for the handle

inc/compatibilities/sassy_social_share.php
```diff
@@ -0,0 +1,65 @@
...
+	function should_load() {
+		include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+
+		if ( ! is_plugin_active( 'sassy-social-share/sassy-social-share.php' ) ) {
+			return false;
+		}
+		$ss_options = get_option( 'heateor_sss' );
+		$ss_bars = ['vertical_re_providers', 'horizontal_re_providers'];
+		foreach ( $ss_bars as $key => $bar ) {
+			foreach ( $ss_options[ $bar ] as $index => $value ) {
```
selul: we need to check if $ss_options is an array and if $ss_options[ $bar ] exists, otherwise it will throw a notice

inc/compatibilities/sassy_social_share.php
```diff
@@ -0,0 +1,69 @@
...
+	 */
+	function should_load() {
+		include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+
+		if ( ! is_plugin_active( 'sassy-social-share/sassy-social-share.php' ) ) {
+			return false;
+		}
+		$ss_options = get_option( 'heateor_sss' );
+		$ss_bars = ['vertical_re_providers', 'horizontal_re_providers'];
+		if ( ! empty( $ss_options ) && is_array( $ss_options ) ) {
```
selul: you can return early
```suggestion
		if ( ! is_array( $ss_options ) ) {
		     return false;
		}
```

inc/compatibilities/sassy_social_share.php
```diff
@@ -0,0 +1,69 @@
...
+			foreach ( $ss_bars as $key => $bar ) {
+				if ( array_key_exists( $bar, $ss_options ) ) {
```
selul: we can use isset which is faster and bail early if does not exists. 
```suggestion
				if ( ! isset( $bar[$ss_options] ) ) {
				    continue;
				}
```

Issue #197, opened 2019-12-12
cli wrapper for common operations 


selul:
we can build a simple cli wrapper for the common Optimole plugin operations like 
 - connect to service
 - disconnect
 - enable/disable replacement
 - enable/disable lazyload  
 - update quality
 - enable/disable generic placeholder
 
They can all be under the optimole `namespace`

PR #208, opened 2020-01-08
chore: added cli commands to control settings #197

selul:


selul:
@preda-bogdan seems like the linter is failing

PR #213, opened 2020-01-17
Fix/divi notice logic

GrigoreMihai:
Fixes the notice being displayed on older versions of divi that do not have static css option. 

https://secure.helpscout.net/conversation/1057354236/213565?folderId=2353196

@preda-bogdan


selul:
@GrigoreMihai can you use the latest dev branch, it seems like this one adds some unnecessary commits.

GrigoreMihai:
@selul did it here #217

PR #214, opened 2020-02-05
Deactivate uninstall cleanup

GrigoreMihai:
@preda-bogdan I think it is ok


inc/main.php
```diff
@@ -69,6 +69,9 @@ public function __construct() {
 	public static function instance() {
 		if ( null === self::$_instance ) {
 			add_filter( 'themeisle_sdk_products', array( __CLASS__, 'register_sdk' ) );
+			add_filter( 'ti_custom_deactivate_cleanup', function ($key) {
```
preda-bogdan: Can we have separate functions just to keep everything the looking the same instead of anonimus functions?

optimole-wp.php
```diff
@@ -76,6 +76,11 @@ function optml() {
 
 		return null;
 	}
+	add_action( 'sdk_deactivate_cleanup', function( $shouldCleanup ) {
```
preda-bogdan: same here

optimole-wp.php
```diff
@@ -45,6 +45,10 @@ function optml_autoload( $class ) {
  */
 function optml_deactivate() {
 	if ( is_plugin_active( 'optimole-wp/optimole-wp.php' ) ) {
+		$option_name = array('optml-version', 'optml_dismissed_conflicts', 'optml_settings', 'optml_notice_optin');
+		foreach ( $option_name as $index => $option ) {
+			delete_option( $option );
+		}
```
preda-bogdan: Can we have this in a separate function since we also use the same block of code in the `uninstall.php` ?
It would be easier to maintain if changes are required.

uninstall.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
+	die;
+}
+
+$option_name = array('optml-version', 'optml_dismissed_conflicts', 'optml_settings', 'optml_notice_optin');
+foreach ( $option_name as $index => $option ) {
+	delete_option( $option );
+}
```
preda-bogdan: Call the function that I mentioned earlier.

preda-bogdan:
@GrigoreMihai can you please check that all tests are passing so that we can maybe have this in a future release?

Thank you!

PR #218, opened 2020-02-13
Compatibility/shareaholic

GrigoreMihai:
Fix for: https://secure.helpscout.net/conversation/1080759370/216305/


inc/compatibilities/shareaholic.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+/**
+ * Class Optml_sassy_social_share.
```
selul: there is a typo here

inc/compatibilities/shareaholic.php
```diff
@@ -0,0 +1,54 @@
...
+
+	/**
+	 * Register integration details.
+	 */
+	public function register() {
+		add_action(
+			'wp_enqueue_scripts',
+			function () {
+				wp_register_script( 'optml-typekit', false );
+				wp_enqueue_script( 'optml-typekit' );
```
selul: you use the same handle as [here](https://github.com/Codeinwp/optimole-wp/blob/master/inc/compatibilities/sassy_social_share.php#L49), maybe we can use something unique

PR #220, opened 2020-03-23
Feat/cache token

selul:


assets/vue/components/general.vue
```diff
@@ -73,6 +89,9 @@
             this.$emit('update-status', !this.isReplacerOff);
         },
         methods: {
+	        clearCache: function () {
+		        this.$store.dispatch('clearCache', { token: this.site_settings.cache_buster });
```
selul: we can avoid sending the cache token.

assets/vue/store/actions.js
```diff
@@ -97,6 +97,30 @@ const disconnectOptimole = function ( {commit, state}, data ) {
...
+const clearCache = function ( {commit, state}, data ) {
+	commit( 'toggleLoading', true );
+	return Vue.http(
+		{
+			url: optimoleDashboardApp.root + '/clear_cache',
+			method: 'POST',
+			headers: {'X-WP-Nonce': optimoleDashboardApp.nonce},
+			emulateJSON: true,
+			body: {
+				'token': data.token
```
selul: we can avoid sending token.

inc/image.php
```diff
@@ -52,6 +52,13 @@ class Optml_Image {
 	 */
 	private $source_url = '';
 
+	/**
+	 * Cache Buster passed from  Optimole Service.
+	 *
+	 * @var string Active Cache Buster
+	 */
+	private $cache_buster = '1';
```
selul: we can use cache_buster value on user who cleared cache, for rest of them use the path token.

PR #222, opened 2020-03-26
fix: added chines characters to list in order to match urls

preda-bogdan:
Here is the regex test for this change: https://regex101.com/r/Lu1MZd/1


selul:
we need also a unit test for this

preda-bogdan:
@selul I've changed the character list and added the test for chinese.
Let me know if you have any concerns about this.

Thank you!

PR #235, opened 2020-05-11
Split scale lazyload

GrigoreMihai:
This needs https://github.com/Codeinwp/optimole_js_lib/pull/74 to work properly.


inc/admin.php
```diff
@@ -112,8 +113,8 @@ public function inline_bootstrap_script() {
 						s.src = "%s/v2/latest/optimole_lib" + v  + "%s.js"; 
 						b.appendChild(s);
 						w.optimoleData = {
-							lazyloadOnly: "optimole-lazy-only",
 							backgroundReplaceClasses: [%s],
+							lazyloadOnly: %s,
```
preda-bogdan, line 117: I think it would be better to leave this as it was since there might be instances where this is used and instead use a flag for scale eg. `scalingEnabled` that is `bool`

inc/settings.php
```diff
@@ -25,6 +25,7 @@ class Optml_Settings {
 		'max_width'            => 2000,
 		'admin_bar_item'       => 'enabled',
 		'lazyload'             => 'disabled',
+		'scale'                => 'disabled',
```
preda-bogdan, line 28: This should be enabled by default as right now the option is enabled as part of the lazyload.
If we disable it it will disable rescale on all active websites until the option is toggled.

PR #236, opened 2020-05-12
chore: toggles for css/js minify

preda-bogdan:
References: Codeinwp/optimole-service#159


inc/asset.php
```diff
@@ -33,12 +40,20 @@ class Optml_Asset extends Optml_Resource {
 	public function __construct( $url = '', $args = array(), $cache_buster = '' ) {
 		parent::__construct( $url, $cache_buster );
 
+		$settings = new Optml_Settings();
```
selul, line 43: we can have some static props on the class for the css/js minify status rather than getting them each time we do an asset process.

inc/asset_properties/minify.php
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	/**
+	 * Set property value.
+	 *
+	 * @param mixed $value Value to set.
+	 */
+	public function set( $value ) {
+
+		if ( $value === 'auto' ) {
```
selul, line 40: minify should support only m:0, m:1, or missing.

inc/asset_properties/minify.php
```diff
@@ -0,0 +1,73 @@
...
+	 * Return property value.
+	 *
+	 * @return mixed
+	 */
+	public function get() {
+		return $this->minify;
+	}
+
+	/**
+	 * Return ImageProxy URL formatted string property.
```
selul, line 66: the comment is inherited from the image asset

PR #237, opened 2020-05-21
Feat/report route

preda-bogdan:


GrigoreMihai:
@preda-bogdan I think it should be ok now

inc/rest.php
```diff
@@ -62,6 +62,19 @@ public function register() {
...
+			array(
+				array(
+					'methods'             => \WP_REST_Server::EDITABLE,
+					'permission_callback' => function () {
+						return current_user_can( 'manage_options' );
+					},
+					'callback'            => array( $this, 'check_redirects' ),
+				),
+			)
+		);
```
preda-bogdan, line 77: I think we can also have some basic validation here or the arguments

inc/rest.php
```diff
@@ -608,5 +621,67 @@ public function update_option( WP_REST_Request $request ) {
...
+				}
+			}
+		}
+		return new WP_REST_Response(
+			array(
+				'status' => $status,
+				'log'   => $result,
+			),
+			200
+		);
```
preda-bogdan, line 684: I think we can use the built in method for this class `$this->response()`

preda-bogdan:
Small changes other than that it looks good I just need to test it.

inc/admin.php
```diff
@@ -33,6 +33,9 @@ public function __construct() {
 		add_action( 'admin_notices', array( $this, 'add_notice' ) );
 		add_action( 'admin_notices', array( $this, 'add_notice_upgrade' ) );
 		add_filter( 'admin_body_class', array( $this, 'add_body_class' ) );
+		if ( ! is_admin() && $this->settings->is_connected() ) {
```
selul, line 36: we need to have this enabled only if the replacer is ON

assets/js/report_script.js
```diff
@@ -0,0 +1,74 @@
...
+					switch (data.code) {
+						case  "deactivated" : {
+							optmlAdmin.innerHTML = "<li><div class='ab-item ab-empty-item'>Your account is permanently disabled<div><li>";
+							break;
+						}
+						case  "log" : {
+							let notice = "";
+							for (let domain of Object.keys(data.data)) {
+								if (data.data[domain] === "whitelist") {
+									notice += "<li><div class='ab-item ab-empty-item'> The domain: " + domain + " is not added to the whitelist <div><li>";
```
selul, line 57: we need to have the strings translatable, better to send the error messages via the rest endpoint rather than using them in the js

PR #257, opened 2020-07-14
Feat/lazyload video

### LINKED ISSUE #253: lazyload youtube videos and investigate youtube embeds optimizations
**Is your feature request related to a problem? Please describe.**
We can investigate a way of lazy loading the youtube videos embedded in a page.

This can be a toggle setting inside the lazyload tab.

GrigoreMihai:
Fixes #253 

Needs https://github.com/Codeinwp/optimole_js_lib/pull/76 to work.


preda-bogdan:
One note here, is that we remove the `src` attribute and I think this is not valid for the `iframe` element. https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element

We could have `src="about:blank"`

inc/admin.php
```diff
@@ -81,6 +81,7 @@ public function inline_bootstrap_script() {
 		$default_network       = ( $this->settings->get( 'network_optimization' ) === 'enabled' );
 		$retina_ready          = ! ( $this->settings->get( 'retina_images' ) === 'enabled' );
 		$scale_is_disabled     = ( $this->settings->get( 'scale' ) === 'enabled' );
+		$video_lazy_placeholder = OPTML_URL . 'assets/img/optml_loading.gif';
```
selul, line 84: here to avoid an extra request, can't we use an SVG loading icon inline? or a base64 encoded image.

GrigoreMihai: yes, I'm looking for one to add now

GrigoreMihai:
> One note here, is that we remove the `src` attribute and I think this is not valid for the `iframe` element. https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element
> 
> We could have `src="about:blank"`, this should be more compliant and should behave the same.

I added it since it seems good for compatibility with all browsers and situations, but from the docs it seems like it could be empty and browsers should assume src to be `about:blank`

PR #261, opened 2020-07-22
Feat/support native lazyload

GrigoreMihai:


GrigoreMihai:
I have added here the speed tests to compare native lazyload with classic one. 

https://docs.google.com/document/d/10GFyT0sTWnl-hY-whwZZJS--rT9QrzPouW4H_vEe1nA/edit?usp=sharing

Seems like the native one is just a bit faster from time to time.

PR #280, opened 2020-08-31
chore: excluded iframes nested inside noscript from lazyload

### LINKED ISSUE #278: Processing `iframe` inside `noscript` tags
**Describe the bug**
When `iframe` lazy-loading is active and there is an `iframe` defined inside `noscript` we also process those.
https://wordpress.org/support/topic/lazyloading-iframes-issue/

**To Reproduce**
1. Create an `iframe` inside a `noscript` tag
2. Activate lazy-loading for iframes inside Optimole
3. See that the iframe from `noscript` also gets processed.

**Expected behavior**
Should not process existing `iframe` inside `noscript`

GrigoreMihai:
I have added the modified regex here for easier testing in case you want to check it. https://regex101.com/r/ijlQP2/5

Fixes #278.

### DIFF
--- a/inc/lazyload_replacer.php
+++ b/inc/lazyload_replacer.php
@@ -259,11 +259,14 @@ final class Optml_Lazyload_Replacer extends Optml_App_Replacer {
 	 */
 	public function lazyload_video_replace( $content ) {
 		$video_tags = array();
-		preg_match_all( '#<iframe(.*?)></iframe>#is', $content, $video_tags );
+		preg_match_all( '#(?:<noscript\s*>(\s*|.*))?<iframe(.*?)></iframe>#is', $content, $video_tags );
 
 		$search = array();
 		$replace = array();
 		foreach ( $video_tags[0] as $video_tag ) {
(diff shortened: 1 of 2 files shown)


inc/lazyload_replacer.php
```diff
@@ -259,11 +259,14 @@ public function lazyload_tag_replace( $new_tag, $original_url, $new_url, $optml_
 	 */
 	public function lazyload_video_replace( $content ) {
 		$video_tags = array();
-		preg_match_all( '#<iframe(.*?)></iframe>#is', $content, $video_tags );
+		preg_match_all( '#(?:<noscript\s*>(\s*|.*?))?<iframe(.*?)></iframe>#is', $content, $video_tags );
 
 		$search = array();
 		$replace = array();
 		foreach ( $video_tags[0] as $video_tag ) {
+			if ( strpos( $video_tag, '<noscript' ) !== false ) {
```
selul, line 267: @GrigoreMihai we should the flag in the filter from should_lazyload_iframe

selul (changes requested):

PR #337, opened 2021-02-12
chore: updated elementor background lazyload classes

### LINKED ISSUE #331: review elementor compatibility
**Describe the bug** 

We need to review the elementor compatibility with the most recent version. We need to double-check the background lazyload compatibility and CSS URL replacement making sure that they work as before.

If they no work anymore, we need to have version-based compatibility, i.e for the old Elementor version keep the same mechanism/CSS classes, for a new one, add the latest version.

Not sure if we can build some nightly tests where we test all the compatibilities in some docker environment and Cypress or similar.

GrigoreMihai:
Fixes #331 . 

There is only one more way to add background that was not covered by the previous compatiblity (the previous classes are working as before). The css replacement works the same in this version. 

For testing I think I can update the current elementor version on testing.optimole.com and include the new background in the cypress test ?

### DIFF
--- a/cypress/integration/test_elementor_background_lazyload.js
+++ b/cypress/integration/test_elementor_background_lazyload.js
@@ -9,6 +9,13 @@ describe("Check Elementor Background Page", function () {
       .should("have.attr", "class")
       .and("include", "optml-bg-lazyloaded");
   });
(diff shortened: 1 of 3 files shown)


selul:
@GrigoreMihai sure.

PR #339, opened 2021-02-18
chore: fix bundle and cypress lint packages conflict

GrigoreMihai:
This will fix the js bundle generation while keeping the cypress lint as it is now. @gutoslv Let me know what you think


package.json
```diff
@@ -45,26 +44,22 @@
...
-    "eslint-plugin-html": "^6.1.1",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-prettier": "^3.3.1",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-vue": "^7.5.0",
-    "eslint-webpack-plugin": "^2.5.0",
+    "eslint": "^4.19.1",
+    "eslint-config-standard": "^10.2.1",
+    "eslint-loader": "^1.9.0",
```
gutoslv, line 49: This package is deprecated, we should update it later. It will also need to migrate webpack to version >4 to use the eslint-webpack-plugin.

gutoslv (approved):
These changes will be temporary while we don't upgrade webpack to version >4.x.x in order to use eslint-webpack-plugin over eslint-loader which is deprecated. For now, it's a workaround so we can still build it while having eslint rules for Cypress tests.

PR #352, opened 2021-03-29
Feat/multiple app support

preda-bogdan:


preda-bogdan:
References: Codeinwp/optimole-service#413
Notes: I think I forgot to add something or the response from the service changed. I've made some changes and tested the following:

1. Automatic app selection when only one app is available.
2. Multiple apps available case has a 2 step process.
3. The domain is whitelisted upon connecting for both cases and the domain is assigned to the correct application.

PR #357, opened 2021-04-19
chore: using date when saving image

GrigoreMihai:


inc/media_offload.php
```diff
@@ -443,6 +443,10 @@ public function rollback_and_update_images( $image_ids ) {
 			}
 
 			$get_url = json_decode( $get_response['body'], true )['getUrl'];
+			$date = null;
+			if ( isset( json_decode( $get_response['body'], true )['date'] ) ) {
```
selul, line 447: try to make the code DRY, there are 3 places where you use json_decode for the same object. Also, you need to check if the decoded string is an object, in case that malformed response might get there

selul (changes requested):

PR #361, opened 2021-04-27
chore: added latin unicode characters to match url

### LINKED ISSUE #326: bug: Some special characters break the URL
**Describe the bug**
Similar to how some special language characters were not being processed correctly.
Reported here:
https://secure.helpscout.net/conversation/1409481589/263853?folderId=2353196

**To Reproduce**
1. Add an image that contains inside the file path a special char like ©
2. The image will break

**Expected behavior**
The image should load.

GrigoreMihai:
Fixes #326.


gutoslv:
The image with special character is not being optimized by optimole. You can check it here: https://residential-jambu.jurassic.ninja/wp-admin

User: demo 
pwd: sxRKEtaRdmhZ

GrigoreMihai:
> The image with special character is not being optimized by optimole. You can check it here: https://residential-jambu.jurassic.ninja/wp-admin
> 
> User: demo
> pwd: sxRKEtaRdmhZ

On this example you are using a character that is out of the range we added with this fix. The range we support now is: u+0080 - u+017F, you can see it here: https://www.utf8-chartable.de/unicode-utf8-table.pl .

gutoslv (approved):
Everything is working as expected. The images with emoji in URL are not being optimized by optimole, but it does not break. Images with special characters in URL are being optimized.

Issue #360, opened 2021-04-26
Consider adding decoding=async to img tags


selul:
We can look at how using the decoding=async attribute to img tags will help us here. We need to evaluate the impact using the generic placeholder/without generic placeholder, with native lazyload/without native lazyload. 

This should auto-applied to all image tags, independent of lazyload mechanism.

Issue #355, opened 2021-04-19
exclude first X number of images from a certain page from lazyload


selul:
A good thing that we could add in the plugin is the ability to exclude first X images from the pages from lazyload, assuming that those images are in the fold. 

The default could be set to 3 and we can offer a setting inside the Lazyload tab to customize this. 

The implementation can assume an internal buffer index which can count the applied lazyload on images and block it on index < desired limit.

ineagu:
a bit unrelated note @selul having a service-connected to webpagetest/gtmetrix that check all the possible settings and pick the fastest config, might be interesting, if not too complex.

I guess we can already test the value of this by offering $199 one-time manual configuration by us for best results and see if anyone cares for auto-config.

PR #362, opened 2021-04-29
Feat/lazyload video tags

### LINKED ISSUE #351: research on video lazyload feature
**Is your feature request related to a problem? Please describe.**
Right now we do lazyload on iframe-based videos, like youtube or similar but we ignore the video tags. We can look if adding a lazyload mechanism to the video tag will add any benefit.  

**Describe the solution you'd like**
Lazyload the video tag as we do when we have the GIF->Video conversion on. 

**Additional context**
https://secure.helpscout.net/conversation/1456879818/271316?folderId=2353196

GrigoreMihai:
Fixes #351 , #355, #360 .

I will add below some pictures for `decoding='async'` tests. From what I see there is not a big difference. (I think the browser already picks what is the best and it might usually pick async and that is why we do not see a difference)

Decoding async: 

no_lazyload:

no_lazyload_decoding

lazyload:

lazyload_decoding

lazyload_generic:

lazyload_generic_decoding

native_lazyload: 

lazyload_native_decoding

No decoding:

no_lazyload:

no_lazyload

lazyload:

lazyload

lazyload_generic:

lazyload_generic

lazyload_native:

lazyload_native


inc/tag_replacer.php
```diff
@@ -210,11 +217,16 @@ public function process_image_tags( $content, $images = [] ) {
 			);
 
 			// If the image is in header or has a class excluded from lazyload, we need to do the regular replace.
-			if ( $images['in_header'][ $index ] ) {
+			if ( $images['in_header'][ $index ] || self::$lazyload_skipped_images < $this->settings->get( 'skip_lazyload_images' ) ) {
```
selul, line 220: The index check is not properly placed, it should be inside `can_lazyload_for` as last condition since `regular_tag_replace` from within `lazyload_tag_replace`

inc/tag_replacer.php
```diff
@@ -210,11 +217,16 @@ public function process_image_tags( $content, $images = [] ) {
...
 				$image_tag = $this->regular_tag_replace( $image_tag, $images['img_url'][ $index ], $new_url, $optml_args, $is_slashed, $tag );
 			} else {
 				$image_tag = apply_filters( 'optml_tag_replace', $image_tag, $images['img_url'][ $index ], $new_url, $optml_args, $is_slashed, $tag );
 			}
+			if ( strpos( $image_tag, 'decoding=' ) === false ) {
+				$pattern = '/img(.*?)/is';
+				$replace = $is_slashed ? addcslashes( 'img decoding="async" $1', '/' ) : 'img decoding="async" $1';
```
selul, line 227: we can use `decoding=async` too, to avoid extra quotes.

selul (changes requested):

selul (changes requested):
Also we need phpunit tests for each one of the features

inc/tag_replacer.php
selul, line 220: Also, we need to make the limit filtrable for users to be able to adjust this dynamically per context, a filter like `optml_lazyload_images_skip` should be fine.

selul (changes requested):

inc/lazyload_replacer.php
```diff
@@ -347,7 +366,9 @@ public function can_lazyload_for( $url, $tag = '' ) {
 		if ( defined( 'OPTML_DISABLE_PNG_LAZYLOAD' ) && OPTML_DISABLE_PNG_LAZYLOAD ) {
 			return $type['ext'] !== 'png';
 		}
-
+		if ( Optml_Tag_Replacer::$lazyload_skipped_images < apply_filters( 'optml_lazyload_images_skip', $this->settings->get( 'skip_lazyload_images' ) ) ) {
```
selul, line 369: I would use an approach similar to the rest of the filterable settings, one example is get_lazyload_bg_classes to avoid applying the filter each time we call `can_lazyload_for`

selul (changes requested):

selul (approved):

gutoslv:
I didn't manage to make video tags lazyload, at least the tag didn't appear. May you check it? Here is the [video](https://www.loom.com/share/23db3cc6b64849a9a95cdf0a1894f394)

And the instance is: https://technological-guineafowl.jurassic.ninja

user: demo
password: FQQZYF0GSIlu

gutoslv:
@GrigoreMihai I've created a new instance since that one is dead, with the plugin from this PR it's not showing the lazyload tag as it shows on images, but it looks like it is lazyloading the videos. 

Please take a look at here:
https://safe-deer.jurassic.ninja/wp-admin/
user: demo
pwd: NFD3yBPyr^$eWzHVWedBrysk

If the behavior is to be without the lazyload tag, then it's working.

GrigoreMihai:
> @GrigoreMihai I've created a new instance since that one is dead, with the plugin from this PR it's not showing the lazyload tag as it shows on images, but it looks like it is lazyloading the videos.
> 
> Please take a look at here:
> https://safe-deer.jurassic.ninja/wp-admin/
> user: demo
> pwd: NFD3yBPyr^$eWzHVWedBrysk
> 
> If the behavior is to be without the lazyload tag, then it's working.

It just needs to have the option enabled. You can check now I enabled it and it looks ok. 

Screenshot 2021-05-10 at 11 11 08

gutoslv:
It's working fine. Moved to ready to merge.

PR #364, opened 2021-05-11
chore: auto connecting if api key is defined

### LINKED ISSUE #314: extend wp-config configuration for api key
**Describe the bug**
Right now we cover a large variety of settings setup using wp-config constants but we are missing an important one for allowing Optimole to auto-connect. 

When using the env mode, we can have a wp-config constant for API key. When defined and if the site is not connected we auto-connect to the service on an init hook or something.   

This should help us automate the Optimole plugin setup.

GrigoreMihai:
Fixes #314 , https://github.com/Codeinwp/optimole-service/issues/387.


gutoslv:
I've managed to test it on the staging site and it's working well. What are the test instructions for the plugin?

GrigoreMihai:
> I've managed to test it on the staging site and it's working well. What are the test instructions for the plugin?

Checking if the plugin will auto connect when the API key is defined and if the other options you can set using environment variables work as before, is what I think that needs to be tested for this. Also, here is the documentation for this if you need it or if you think it can be improved https://docs.optimole.com/article/1316-setting-plugin-options-via-wp-config?preview=5fb3e3624cedfd001610d9fc .

PR #367, opened 2021-05-18
chore: added filter to remove redundant schema

### LINKED ISSUE #346: Allow for urls without the schema part inside the url
This is related to this ticket: https://secure.helpscout.net/conversation/1377308904/259037?folderId=2353196

Add a checkbox option inside the plugin to exclude the schema and domain from the URL if possible.

This will change the url from: 
https://key.i.optimole.com/lhj-JEs.Kj7b~a1c2/w:auto/h:auto/q:auto/https://example.com/path/example_image.jpg

To:
https://key.i.optimole.com/lhj-JEs.Kj7b~a1c2/w:auto/h:auto/q:auto/example.com/path/example_image.jpg

A thing to watch here is if we rely on the schema part to identify domains in our infrastructure, eg. Edge where we tokenize the path.

GrigoreMihai:
Fixes #346 . Needs to use https://gist.github.com/GrigoreMihai/ff2b866b4a91cb31e3117549ed199acc .

### DIFF
--- a/inc/resource.php
+++ b/inc/resource.php
@@ -45,7 +45,7 @@ abstract class Optml_Resource {
 			throw new \InvalidArgumentException( 'Optimole resource builder requires the source url to optimize.' ); // @codeCoverageIgnore
 		}
 		$this->set_defaults();
-		$this->source_url = apply_filters( 'optml_redundant_schema', $url );
+		$this->source_url = apply_filters( 'optml_processed_url', $url );
 
 		$this->cache_buster = $cache_buster;
 	}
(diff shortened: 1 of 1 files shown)


gutoslv:
@GrigoreMihai how may I test it? I've added the gist to the plugin it doesn't show the option.

GrigoreMihai:
> @GrigoreMihai how may I test it? I've added the gist to the plugin it doesn't show the option.

We decided not to make it with an option on the interface, it will work by giving the plugin to the users that want this and once enabled the urls should have only one schema (images and assets). If you enabled the plugin and are also using the plugin from the pr for Optimole urls should have only one http(s) .

gutoslv (approved):

Issue #378, opened 2021-07-05
Warning/Notice appears next to other types of files (mp4, PDF) in media library when offloading is enabled


vytisbulkevicius:
**Describe the bug**
If you have "**Enable offloading images**" enabled in Optimole's settings:
![image](https://user-images.githubusercontent.com/36594177/124486535-5e407100-ddb6-11eb-8aa8-33bdb18c2e96.png)

Next to other type of files (like mp4 or PDF) the Notice/Warning appears in Media Library next to those files (in List view):
![image](https://user-images.githubusercontent.com/36594177/124486700-89c35b80-ddb6-11eb-8dca-9166890ca0df.png)

**To Reproduce**
Have PDF or mp4 file in media library when offloading images is enabled in Optimole.
You should see the warning in media library.

Reported here - https://wordpress.org/support/topic/php-warnings-in-media-library-on-non-image-files/

**Screenshots**
Screenshots provided

**Expected behavior**
No warnings and notices.

**Isolating the problem (mark completed items with an [x]):**
- [X] I have deactivated other plugins and confirmed this bug occurs when only Optimole plugin is active.
- [X] This bug happens with a default WordPress theme active
- [X] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
<details>
 WordPress Version: 5.7.2
 Optimole Version:  3.1.1
 PHP versions: 7.4.20 (Supports 64bit values)
</details>

Issue #373, opened 2021-06-04
Improve the use of plugin rest endpoints


GrigoreMihai:
Right now we are only getting the root endpoint and building the routes from the interface eg: https://github.com/Codeinwp/optimole-wp/blob/master/assets/vue/store/actions.js#L151 . We could make this more reliable by using the `path` argument for the `rest_url` function and building an array with the endpoints. 

Doing this we could possibly avoid cases like: https://secure.helpscout.net/conversation/1517595463/279258/ .

PR #382, opened 2021-08-03
Offload logic change

### LINKED ISSUE #381: Media offload resource usage
With the current version of the plugin we are using to many resources which may cause the offload/rollback to fail depending on the server. This is due to the logic we have ie. for every image we rollback check all posts/pages if they contain the image and replace it. We will change the logic like this: iterate once through the pages/posts and for every post/page get the images  offload/rollback them and then update the post. This should increase the speed of sync/rollback and batching the posts query and the images offload will remove the errors as we will only have small requests to handle at once. To test if the logic change works we can use our staging sites to test the improved version as they are big enough to replicate the issue with current version.

GrigoreMihai:
Fixes #381 , #378 . 

For testing apart from what is in the issues we also need to check:

1. We want to test the performance on big sites (with lots of images and a lot of posts/pages), our staging sites are good for this if you cannot build some big test instances quickly ( it is good to also test on some other instances that our staging sites ).
2. We also want to check that our replacement in the pages works with various contexts ( builders, themes, eg: avada ), the case where it is not intended to work is when the images are present in separate css files. 
3. That the library bulk actions to push or rollback images are working properly ( you can select a maximum of one library page at once and you will be redirected to the plugin interface where a link back to the library will appear when the jobs finishes to sync/rollback the selected images ).
4. That the cli for the same jobs: `wp optimole media move_to_optimole`, `wp optimole media rollback_images` are working properly.
5. We also want to test this on servers with not so much processing power. (jurrasic ninja works if you cannot find something else) 
6. Upload some other files apart from images and check that there are no errors for them ( we only work with images and the plugin should not interfere with other file types )
7. We can also look at the time estimate that was added to the jobs to be sure it is accurate (it is normal for it to change while the job runs (increase or decrease), but it should not for example display a few hours and then the job to end ).


gutoslv:
@GrigoreMihai I've found an issue with the estimated time: https://vertis.d.pr/47TVli It stops updating after a time.

You can check it on this instance: 
https://western-trout.jurassic.ninja/wp-admin
user: demo
pwd: 7tMORXEkNXJA

GrigoreMihai:
> @GrigoreMihai I've found an issue with the estimated time: https://vertis.d.pr/47TVli It stops updating after a time.
> 
> You can check it on this instance:
> https://western-trout.jurassic.ninja/wp-admin
> user: demo
> pwd: 7tMORXEkNXJA

It was a bug the worker from dev can not process some gif (this part is expected it has low memory) from the site and the plugin code was not skipping failed images and stopped. It should work now I managed to sync everything on that instance after the fix. Also this pr is now also fixing #373 as I merged some other fixes (to test the changes related to that we just need to check that the actions from the plugin work as before eg: connect, disconnect, refresh stats, select app etc.) .

PR #385, opened 2021-08-16
chore: fix woocommerce product page lazyload conflict

### LINKED ISSUE #366: Compatibility woocommerce variations plugins
When using variations for woocommerce products the first image from the list is replaced with the lazyload placeholder, it looks like it is first lazyloaded correctly and then other js script puts the placeholder back. eg: Variation Swatches for WooCommerce, Product Variations Swatches for WooCommerce 

https://secure.helpscout.net/conversation/1501150672/277177?folderId=2353196

GrigoreMihai:
Fixes #366 ( I was not able to replicate the issue ) and also fixes some other issue that I found while testing: when picking a different image from a product gallery it is blurry because we use the smaller size https://vertis.d.pr/oTxVnf . ( happens when using thumbnail size for the product gallery for example )

### DIFF
--- a/inc/compatibilities/woocommerce.php
+++ b/inc/compatibilities/woocommerce.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * Class Optml_elementor_builder
+ *
+ * @reason Adding selectors for background lazyload
+ */
+class Optml_woocommerce extends Optml_compatibility {
+
+
+	/**
+	 * Should we load the integration logic.
+	 *
+	 * @return bool Should we load.
+	 */
+	function should_load() {
+		include_once( ABSPATH . 'wp-admin/includes/plugin.php' );
+		return Optml_Main::instance()->admin->settings->use_lazyload() && is_plugin_active( 'woocommerce/woocommerce.php' );
+	}
+
+	/**
+	 * Register integration details.
+	 */
+	public function register() {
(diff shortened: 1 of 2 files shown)


inc/compatibilities/woocommerce.php
```diff
@@ -0,0 +1,40 @@
...
+	}
+	/**
+	 * Add ignore page lazyload flag.
+	 *
+	 * @param bool $old_value Previous returned value.
+	 *
+	 * @return bool If we should lazyload the page.
+	 */
+	public function check_product_page( $old_value ) {
+		if ( is_product() ) {
```
selul, line 35: This fix looks a bit too loose, can't we exclude exactly the problematic area? Also, we can investigate if there is any custom lazyload that we could do to gain maximum efficiency when using Optimole

selul (changes requested):

inc/compatibilities/woocommerce.php
GrigoreMihai: The problem is on every single product page ( that function is not so intuitive in the name, but it does exactly this check that we need https://docs.woocommerce.com/document/conditional-tags/ ) and because there is not too much to lazyload on the page with just the product it seems a good idea to exclude the page. If we want to keep the lazyload on this pages as they do not include lazyload for the related products for example so it will still gain from ours we could add their lazyload src `data-src` to our flag list when we detect their plugin and this should also give our benefits without interfering with their lazyload ?

selul (approved):

Issue #393, opened 2021-10-04
Ability to clear resources by type (CSS/JS without touching images)


vytisbulkevicius:
**Is your feature request related to a problem? Please describe.**
The client raised this request:
When they edit CSS/JS files, they need to click "Clear cached resources" which regenerate resources in CDN. They have over 10,000 product images, so it would be very helpful if they and other clients can just refresh the CSS/JS cache without regenerating the images in the CDN.

**Describe the solution you'd like**
Maybe we can have a checkbox to check if we want to refresh only CSS/JS or also images next to the button.

**Describe alternatives you've considered**
We can refresh all the resources only.

**Additional context**
Ticket for reference - https://secure.helpscout.net/conversation/1649537403/294607

PR #395, opened 2021-11-01
chore: added specific clear cache option for assets

GrigoreMihai:
Fixes https://github.com/Codeinwp/optimole-wp/issues/393 .


GrigoreMihai:
@irinelenache I added the changes, regarding the empty space in the right before connecting the service it is normal. You can use this build to test it, you can unmark any of the check boxes if they're not properly fixed.

JohnPixle:
@irinelenache thanks for the input.

> The View sample image [button](https://vertis.d.pr/xCESPG) looks pretty simple, more like a link, shouldn't it have some styling or a bigger font size?

Regarding this @GrigoreMihai, the text links should be underlined, the color is correct.
As I see it, it should also be vertically aligned properly in relation to the button.

Let me know if you need any further help with this
Screenshot 2022-02-11 at 13 43 05
!

irinelenache:
@GrigoreMihai Tested again and most of the issues are fixed 🚀 , only the first one still happens for me.
Found something else though:

- [ ] With **Shop** starter site imported, when the image quality is set to 99 or 100, the size percent will appear as bigger than 100% https://vertis.d.pr/6LuRIF . Also, i think the **minus** is not necessary when the word **smaller** is used

GrigoreMihai:
@irinelenache Thanks for the help, I also added the fixes for the latest things let me know if you still see them. The minus was only getting displayed when the optimized size was greater then the original size, should not appear now, also if you have that image you can send it to me.

irinelenache:
@GrigoreMihai Tested and all the issues are fixed now, thank you 🚀

PR #421, opened 2022-04-18
Normalize content url

GrigoreMihai:
Fixes https://github.com/Codeinwp/optimole-service/issues/541 .

### DIFF
--- a/inc/app_replacer.php
+++ b/inc/app_replacer.php
@@ -368,8 +368,13 @@ abstract class Optml_App_Replacer {
 		];
 		$this->upload_resource['url_length'] = strlen( $this->upload_resource['url'] );
 
-		$content_parts = parse_url( content_url() );
+		$site_url = get_site_url();
+		$content_url = content_url();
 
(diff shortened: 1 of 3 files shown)


inc/app_replacer.php
```diff
@@ -368,8 +368,13 @@ public function set_properties() {
 		];
 		$this->upload_resource['url_length'] = strlen( $this->upload_resource['url'] );
 
-		$content_parts = parse_url( content_url() );
+		$site_url = get_site_url();
+		$content_url = content_url();
 
+		if ( strpos( $content_url, $site_url ) === false && strpos( $content_url, '/' ) === 0 ) {
```
selul, line 374: Isn't this condition `strpos( $content_url, '/' ) === 0` enough?

selul (changes requested):

inc/app_replacer.php
GrigoreMihai: yes it is, changed it

selul (approved):

PR #423, opened 2022-04-19
chore: added action to check domain change

GrigoreMihai:
Fixes https://github.com/Codeinwp/optimole-service/issues/542 .

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -34,6 +34,7 @@ class Optml_Admin {
 		add_action( 'admin_notices', [ $this, 'add_notice_upgrade' ] );
 		add_filter( 'admin_body_class', [ $this, 'add_body_class' ] );
 		add_action( 'optml_daily_sync', [ $this, 'daily_sync' ] );
+		add_action( 'init', [ $this, 'check_domain_change' ] );
 		add_action( 'admin_init', [ $this, 'maybe_redirect' ] );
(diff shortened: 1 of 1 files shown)


inc/admin.php
```diff
@@ -42,6 +43,20 @@ public function __construct() {
 
 	}
 
+	/**
+	 * Checks if domain has changed
+	 */
+	public function check_domain_change() {
```
selul, line 49: what happens if the user is not connected?  

Also, I find the naming a bit unrelated, since it has the current domain, not the previous domain. I.e my suggestion would be to have `optml_current_domain` as key.

selul (changes requested):

inc/admin.php
GrigoreMihai: It would not have the data to perform the daily sync, I'll change it to add the action only for connected users. 

That's right the option is always updated to the current domain, will change it. 

I had one question, I was thinking if it would be a better idea to only add the action if the scheduling of the `daily_sync` fails, ie: the response from here https://github.com/Codeinwp/optimole-wp/blob/master/inc/admin.php#L39 is not `true` ? (right now with this change there is no more whitelist redirect on domain change and adding it only if the `daily_sync` fails will still allow redirects until cron triggers that why I'm not sure about this)

irinelenache:
@GrigoreMihai Managed to test the feature and it works fine, didn't find any problem 🚀

PR #436, opened 2022-08-01
Various compatibilities

### LINKED ISSUE #116: hook into divi cache mechanism and replace the urls in the minified files
When the Divi cache is on we are unable to process the images embedded into the css files.

Similar to how we do for beaver and Elementor, we can hook to this process too.

GrigoreMihai:
Fixes #116 .

Things we need to test for this:

- [x] For `beaver builder` we need to test that the css/js files it creates have the image urls optimized
- [x] For `cache enabler` the stored page cache should have optimized urls 
- [x] For `divi` the static css files created with this theme/builder after page creation should have optimized images, for this you need to login with the account mentioned in the handbook and download the latest divi theme version (the handbook zip is outdated), for older versions you will still see the conflict notice and have to manually clear the static files
- [x]  Same for `elementor` we need to check that images stored in css are optimized
- [x] For `slider revolution` we need to make sure that the slider images are not lazyloaded (it will just display blank images if we lazyload there)
- [x]  For `w3 total cache` if using our serve css/js option and their minify css/js then the files optimized are included in the minified file (after following the details in the conflict desc)
- [x] For `elementor, divi` changing the plugin settings (eg: quality) should clear the static files and update the images in the css, same for `cache enabler, w3 total cache` should happen with the images from their caches.
- [x] Also for the plugins from the previous check if you first create the pages and then connect Optimole, simply connecting the plugin should clear the caches/static files and optimize the images in then
- [x] For `yith quick view for woocommerce` the pop-up resulted after using the quick view button should have an optimized image


irinelenache:
@GrigoreMihai Did some testing on this and didn't find any problem on the mentioned points 🚀 Might be worth to check yourself how it behaves in some more complex use cases, because i get a bit lost there 😅

PR #438, opened 2022-08-05
Multiple enhancements

GrigoreMihai:
I added here the changes for one click create account and connect (for this I also added a separate route [here](https://github.com/Codeinwp/optimole-service/pull/595) ), I also made it able to start with a query string: accessing the plugin page after activation like this: `http://optimoletest.test/wp-admin/upload.php?page=optimole&optimole_connect=mihai%2Btest5%40optimole.com` will auto complete the email field and start the account creation and connection. 

The image size part looks like this: https://vertis.d.pr/KWg7P0 , here I'm not sure if I should add a note similar to: 

"If you have images that are no longer cropped after optimization you should add those images sizes here" 

since this is main reason we have this added and also it might be a bit confusing for some users without a description ? 

For the exact url match I think it will be faster and more reliable if we use the path for it, otherwise we will have to consider `http/https` or `www` and also get the domain at that point and merge it with the path , it's not a problem to implement it but since we run the check before any page render it will add to the site speed in my opinion, what do you think ? 

Now it looks like this: https://vertis.d.pr/GCODhH and to exclude `http://optimoletest.test/shop/` for example the user will have to add the path after the domain `/shop/` .


selul:
@GrigoreMihai, the auto registration process is a bit too complex, to goal is to simplify the setup, not just prefill the field. 

What I would do: 
- [ ] Change the email fetching from URL to some option removed once the connection is made. I.e optml_user_email can be. 
- [ ] Once the user loads the page and finds that option, we load the settings screen directly. Also the registration seems not to work right now, tried here https://gurowe.s.themeisle.com/wp-admin Username gurowe Password Dc83DQTY


The image sizes look a bit confusing; once a size registers, it will automatically apply cropping to the defined width/height images. If this is the case, what's the point of having cropping optional since that's why they add the size to make it cropped? @harshitarora-in let me know if you have any thoughts on this. 

The page matches look very good, and the logic is solid🚀

GrigoreMihai:
It needed to use the staging because the route is not deployed to live (I added the constants needed to the instance you shared). 

The query string also starts the connect and opens the plugin main page like this https://vertis.d.pr/3JqDQz . 

I agree that using an option is easier in this case (also avoids the url encoding that's needed). 

We would still need to pop the loading first since it takes a bit for the route to respond or if we want to avoid that and directly show the plugin settings screen once the option is there we would need to call the account creation route from the plugin that will add the option ?

It won't apply cropping unless selected, but you are right it makes more sense to always have the cropped enabled since without cropping that size is redundant for us as the normal resize does not require a registered image size. I think it's better to change it like "Add a new crop image size" and "Your cropped image sizes" and remove all the other crop references from the interface, what do you think ?

harshitarora-in:
@selul All looks good to me. 

Just a few suggestions: should we send pop-up notifications to users who have used up their free allowance by 90%? If so, should we include an offer for a discount on the pro plan of between 5% and 10%? 

A small number of customers are also unable to predict when their free quota will reset again, thus it would be helpful to include information about the day on which it will do so on the setting page. In addition, we may offer these users additional incentives if they subscribe to the pro plan. 

(I have the plugin installed on my own test site, but I'm not sure about the process when the quota is surpassed.)

selul (changes requested):
> I think it's better to change it like "Add a new crop image size" and "Your cropped image sizes" and remove all the other crop references from the interface, what do you think ?

Yes, let's do this.


Also for auto-connect, once we detect the flag, we should jump directly to the loading screen to simplify the setup.

GrigoreMihai:
> > I think it's better to change it like "Add a new crop image size" and "Your cropped image sizes" and remove all the other crop references from the interface, what do you think ?
> 
> Yes, let's do this.
> 
> Also for auto-connect, once we detect the flag, we should jump directly to the loading screen to simplify the setup.



> > I think it's better to change it like "Add a new crop image size" and "Your cropped image sizes" and remove all the other crop references from the interface, what do you think ?
> 
> Yes, let's do this.
> 
> Also for auto-connect, once we detect the flag, we should jump directly to the loading screen to simplify the setup.

Added this changes in the previous commits,the latest build should jump to the loading if the option exists, updated now also the description for image sizes (missed it previously), also added a note to be more clear for users when to add a crop image size https://vertis.d.pr/Qo0Lvi .

selul (approved):

PR #439, opened 2022-08-17
Development

GrigoreMihai:
#### Features
- Allows users to add cropped image sizes from within the plugin settings
- Adds an option to exclude a page path from optimization using exact matching
- Adds the option to create and connect an account with one click
- Adds filter, <code>optml_keep_copyright</code> , to control if image optimization should keep copyright metadata 
#### Fixes
- Enhances compatibility with Beaver builder to optimize images in javascript files 
- Enhances compatibility with Cache enabler to use the latest plugin's filters and clears page cache when Optimole's settings are updated
- Updates compatibility with Divi theme/builder to optimize images in the static css/js files and regenerate those files when Optimole's settings are updated
- Enhances compatibility with Elementor to update the optimized images in the generated css files upon changing Optimole's settings
- Adds the latest lazyload exclusion flags for Slider Revolution
- Enhances W3 Total Cache compatibility to clear the cache when Optimole's settings are updated
- Updates compatibility with YITH WooCommerce Quick View to optimize quick view images


GrigoreMihai:
The parts we still need to test from this are: 

- [ ] The connect and create account option send the email with the account data and at the same time automatically connects the created account
- [ ] The crop images sizes work as expecting. ie. adding a size eg: 500x700 and then in a page setting the same values for an image tags width and height will make the optimized image in that tag to be cropped. (can bee seen by the flag `rt:fill/g:ce` in the URL. 
- [ ] The match path page exclusion works as expecting. usage eg. for `http://optimoletest.test/360-2/` we need to add  
`/360-2/` in order to exclude the page. It should work with nested paths also. eg: `/products/productone/` .

Adds the plugin changes from this list: https://github.com/Codeinwp/optimole-service/issues/593 .

irinelenache:
@GrigoreMihai Tested and overall it seems to work fine, i didn't exactly understand how the cropping feature works and how should i test it, so i'll give it another try after you send some more instructions 👍 

- [ ] I created an account using irinel-mole@themeisle.com and the user will appear like this in the dashboard/plugin https://vertis.d.pr/sTHAYt
- [ ] After activating the plugin on a tastewp instance with **Shop** starter site imported, the images in the shop page have their quality reduced (pretty visible)
Shop page without the plugin: https://vertis.d.pr/yjgabO
Shop page with the plugin: https://vertis.d.pr/dYRDtW

GrigoreMihai:
> -FsXlf5dCEQ

Regarding the quality reduction is not because of any change. It is because the lazyload (scale) feature is enabled and the retina images option is not, so when the images get downscaled to the container's size there is a drop in quality. With retina enabled it looks much better, let me know if I'm missing something here. 

For the crop it works like this: for example on this page https://optimole-plugin.s1-tastewp.com/335-2/ the image's `width ` and `height` attributes are set to some other aspect ratio `400x500` and by default we do not crop that image if the crop size is not registered. This is how it looks without adding a crop size: https://vertis.d.pr/8J4ai3 . After adding the crop size we will crop the image and respect the aspect ratio that is setup `400x500` : https://vertis.d.pr/x4l0SG .

Basically to test this you can add various images with different width, height setup in the `img` tag and then check that adding those particular crop sizes will change how the image displays. You can also check the image URL to see if crop was applied, by looking for `/rt:fill/g:ce/` flag in the URL.

irinelenache:
@GrigoreMihai 
I tried with a wide image that has an original ratio of `1024x461`, i set it to have `width: 500, height 500` in [the editor](https://vertis.d.pr/qizqMR) and i have added both 1024x461 and 500x500 as crop sizes in optimole settings. In frontend, the image won't have the 500x500 ratio and the flag in the URL is `/rt:fill/g:sm/` https://vertis.d.pr/kjFdGW . Let me know if this is intended or i didn't catch how this should work 🙏

GrigoreMihai:
> @GrigoreMihai I tried with a wide image that has an original ratio of `1024x461`, i set it to have `width: 500, height 500` in [the editor](https://vertis.d.pr/qizqMR) and i have added both 1024x461 and 500x500 as crop sizes in optimole settings. In frontend, the image won't have the 500x500 ratio and the flag in the URL is `/rt:fill/g:sm/` https://vertis.d.pr/kjFdGW . Let me know if this is intended or i didn't catch how this should work 🙏

The flag is ok, is because of the `smart gravity` option ( I just named the default one the first time, but this is also valid). The image you mentioned is not getting resized because you get redirect to the original as it has a bigger size than our limit (5000x5000) . Tested with some other image on that page and it works as expected, eg: https://ml4oxyaa0ryq.b.optml.cloud/ffFoR1Q.MtyI~2bbfa/w:700/h:500/q:mauto/rt:fill/g:sm/f:avif/https://optimole-plugin.s1-tastewp.com/wp-content/uploads/2022/08/artur-tumasjan-42l3tjsJGyw-unsplash.jpeg

irinelenache:
@GrigoreMihai I understood now, managed to test it and it works fine 🚀

PR #443, opened 2022-09-06
chore: fixed % sizes parsed as actual size

GrigoreMihai:
Right now we parse tags like this and we add the width as number: 

 
Not sure if this is the best fix, I'll use it temporary to share the build with the user until we look closer.

### DIFF
--- a/inc/tag_replacer.php
+++ b/inc/tag_replacer.php
@@ -263,10 +263,14 @@ final class Optml_Tag_Replacer extends Optml_App_Replacer {
 	 */
 	private function parse_dimensions_from_tag( $tag, $image_sizes, $args = [] ) {
 		if ( preg_match( '#width=["|\']?([\d%]+)["|\']?#i', $tag, $width_string ) ) {
-			$args['width'] = $width_string[1];
+			if ( strpos($args['width'], "%") === false  ) {
+				$args['width'] = $width_string[1];
+			}
 		}
 		if ( preg_match( '#height=["|\']?([\d%]+)["|\']?#i', $tag, $height_string ) ) {
-			$args['height'] = $height_string[1];
+			if ( strpos($args['width'], "%") === false  ) {
+				$args['height'] = $height_string[1];
+			}
 		}
 		if ( preg_match( '#class=["|\']?[^"\']*size-([^"\'\s]+)[^"\']*["|\']?#i', $tag, $size ) ) {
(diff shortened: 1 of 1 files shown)


inc/tag_replacer.php
```diff
@@ -263,10 +263,14 @@ public function can_replace_tag( $url, $tag = '' ) {
...
 	private function parse_dimensions_from_tag( $tag, $image_sizes, $args = [] ) {
 		if ( preg_match( '#width=["|\']?([\d%]+)["|\']?#i', $tag, $width_string ) ) {
-			$args['width'] = $width_string[1];
+			if ( strpos($args['width'], "%") === false  ) {
+				$args['width'] = $width_string[1];
+			}
 		}
 		if ( preg_match( '#height=["|\']?([\d%]+)["|\']?#i', $tag, $height_string ) ) {
-			$args['height'] = $height_string[1];
+			if ( strpos($args['width'], "%") === false  ) {
```
selul, line 271: it should be height here.

selul (changes requested):

inc/tag_replacer.php
GrigoreMihai: thanks for noticing

selul:
> Right now we parse tags like this and we add the width as number:
> 
> ```
> 
> ```
> 
> Not sure if this is the best fix, I'll use it temporary to share the build with the user until we look closer.

seems like the attribute value is invalid as it only supports numeric ones https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img; we can have a validation routine that use that value if is a number, otherwise, leave it as it is. 

we can also mention this to the user.

GrigoreMihai:
> > Right now we parse tags like this and we add the width as number:
> > ```
> > 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not sure if this is the best fix, I'll use it temporary to share the build with the user until we look closer.
> 
> seems like the attribute value is invalid as it only supports numeric ones https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img; we can have a validation routine that use that value if is a number, otherwise, leave it as it is.
> 
> we can also mention this to the user.

That is a good point I thought it was valid, I already sent the reply before seeing this , but I'll let him know after he replies.

selul (approved):

PR #446, opened 2022-10-06
Development

GrigoreMihai:
* Improved images offload speed by marking processed pages
* Adds conflict validation before the rollback process
* Updated the logging for images that fail when offloading


GrigoreMihai:
@irinelenache I will note here the things that we need to test for this: 

- [ ] Now on the offload option if you stop it in the middle at the next try it knows to pick up from where it left, this can be seen by the timing being lower at the next try (compared to the initial calibration from the first try). The easiest way to see this is: after the offload finishes if you use it again it finishes instantly (before this fix it would loop again through the pages). The same logic is applying for the rollback. 
- [ ] Other thing to test will be the following scenario: you add one image to some page, then you offload all images, after that temporary disable the media cloud feature and add a new image to the page (this image is not offloaded now). Then enable back the media cloud feature and offload the image. The image should be properly offloaded on the page. (this is will test that the marking we do on the pages at the above step is removed when the page is edited and updated by the user)
- [ ] When you use the rollback option if you have installed plugins that can cause any conflicts you will see this notice: https://vertis.d.pr/GirRgP . You can dispel the warning but if you do not disable the plugins mentioned it will appear again not letting you to rollback until you disable the plugins. (eg plugins: Smush, WP optimize, Imagify) 

Let us know here how it goes, when you have the time to check this, thank you.

irinelenache:
@GrigoreMihai Tested and found a single problem, i don't know if i correctly followed the steps for this one, let me know what you think 👀 

- [ ] I installed and activated Smush plugin and didn't get any notice when offloading/rollbacking images in Optimole. I also tried "Smushing" the images and then offloading them with Optimole, and still no notice. 

You can check the behaviour on this instance:
```
 Admin area URL: https://mole.s4-tastewp.com/wp-admin 
 Username: irinel 
 Password: 3B37YkzcZF8 
```

GrigoreMihai:
> 3B37YkzcZF8

The notice is only for rollback, I checked now and smush wasn't active, after activating it I get the notice as expected. https://vertis.d.pr/WildoF , I think it was just that plugin not being active, let me know how it goes, thank you.

irinelenache:
@GrigoreMihai Checked again and the notice is visible and works as expected, thank you 🚀

PR #455, opened 2022-11-21
Fix/wpml compatibility

GrigoreMihai:


GrigoreMihai:
For testing this we need to have an instance with images attached to pages and unattached. Then install the WPML plugin and translate some pages and after that translate the images: https://vertis.d.pr/i/4XHBDk . Then enable the option to offload the images and sync them. With the fix the images should be offloaded. We also need to test that the rollback is working as expected in this context.

irinelenache:
@GrigoreMihai Tested the fix and everything works fine on my end 🚀

selul (approved):

PR #458, opened 2022-12-20
chore: fixed offload deduplication issue

GrigoreMihai:
Fixes https://github.com/Codeinwp/optimole-service/issues/642 .


irinelenache:
@GrigoreMihai Tested and the issue is fixed now 🚀

inc/media_offload.php
```diff
@@ -192,11 +198,49 @@ public static function instance() {
...
+
+		if ( self::$should_not_deduplicate === true ) {
+			return $filename;
+		}
+		$sanitized_filename = str_replace( $ext, '', $filename );
+		$args = [
+			'post_type' => 'attachment',
+			'post_mime_type' => 'image',
+			'post_name__like' => $sanitized_filename,
+			'posts_per_page' => -1,
```
selul, line 229: Why this is required ? On a big environment this will defintely crash.

selul (changes requested):

inc/media_offload.php
GrigoreMihai: need to get the count of images with the same name to append the number like `-1` at the end to fix the issue with images that are uploaded with the same name. The default WP deduplication is using the local files from the uploads to check but because we remove the images that no longer works. Our only way to mimic this is using the database.

inc/media_offload.php
```diff
@@ -192,11 +218,83 @@ public static function instance() {
...
+			do_action( 'optml_log', $data );
+		}
+		if ( ! empty( $data['guid'] ) ) {
+			$filename = wp_basename( $data['guid'] );
+			$ext = $this->get_ext( $filename );
+			$sanitized_post_name = str_replace( '-' . $ext, '', $data['post_name'] );
+		}
+		if ( ! empty( $data['post_name'] ) && $data['post_title'] !== $sanitized_post_name ) {
+			$to_replace_with = $sanitized_post_name . '.' . $ext;
+			$data['guid'] = str_replace( $filename, $to_replace_with, $data['guid'] );
```
selul, line 282: @GrigoreMihai this part seems to repeat pretty often in this PR. We can make it more DRY by moving it into a utility function.

selul (changes requested):

selul (approved):

irinelenache:
@GrigoreMihai Checked again with the latest changes and didn't find any problem ✅

PR #463, opened 2023-01-16
chore: updated iframe lazyload flags to include the default flags #461

### LINKED ISSUE #461: The generic skip-lazy class doesn't work with Iframes.
### Description -

If you have an iframe lazyloaded by optimole and you want to exclude it from lazyload by adding the generic **class ="skip-lazy"**, it no longer excludes the Iframes from the lazyloading.  

The other classes like 'gform_ajax_frame', 'recaptcha', 'original-src', works fine, and if you will use them they prevent the iframes from being lazyloaded.

### Reference - 
reported here - https://secure.helpscout.net/conversation/2120809479/351568?folderId=2353196

image

image

### Technical Details -

WP version - 6.1.1
Optimole Plugin version - 3.5.4

GrigoreMihai:
Fixes #461.

### DIFF
--- a/inc/lazyload_replacer.php
+++ b/inc/lazyload_replacer.php
@@ -323,10 +323,10 @@ final class Optml_Lazyload_Replacer extends Optml_App_Replacer {
 		$search = [];
 		$replace = [];
 		foreach ( $video_tags as $video_tag ) {
-            do_action('optml_log', 'here');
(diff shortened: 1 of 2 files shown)


irinelenache:
@GrigoreMihai Tested and the issue is fixed now, thank you 🚀

PR #464, opened 2023-01-17
chore: add str_starts_with alternative for all php versions

GrigoreMihai:


GrigoreMihai:
The ticket where it was reported and tested: https://secure.helpscout.net/conversation/2122749686/351831?folderId=2353201 .

PR #466, opened 2023-01-20
chore: improve video lazyload regex and preload value

### LINKED ISSUE #467: Videos missed from lazyload
If a video tag uses the `<source>` tag added on a newline then the video tag is ignored from the lazyload. 

Html example: 

```html 
<video controls="">
<source type="video/mp4" src="https://domain.com/video.mp4">
</video>
```

Expected behaviour: the video tag should have a `preload` attribute, like this: 

```html 
<video controls="" preload="metadata">
<source type="video/mp4" src="https://domain.com/video.mp4">
</video>
```

Ticket: https://secure.helpscout.net/conversation/2127450088/352569?folderId=2353201 .

GrigoreMihai:
Fixes the video lazyload regex not catching video tags were the `<source>` tag is added on a newline. Regex before: https://regex101.com/r/DXf5GM/1 and after: https://regex101.com/r/5Vi26p/1 . 

Changes the `preload` value to `metadata`  for a nice placeholder view.  With the previous value `none` it looks empty before opening: https://vertis.d.pr/i/mhm6th . 

It also removes adding redundant `<noscript>` tag when the video uses the `preload` attribute. 
 
Fixes #467.


selul:
@GrigoreMihai let's also create an issue to reference this in testing.

selul (dismissed):

irinelenache:
@GrigoreMihai Can you add some clear test instructions please? 🙏  I tried to test this using the info in the issue and the ticket but i don't exactly get it

GrigoreMihai:
> @GrigoreMihai Can you add some clear test instructions please? 🙏 I tried to test this using the info in the issue and the ticket but i don't exactly get it

Sure: 

- [ ] We need to check that if inside a `<video>` tag the source tag is added on a newline then that tag gets lazyloaded. (ie. we need to check that it has the `preload="metadata"` attribute ). Previously this kind of tags (there is an example in the issue) were not being lazyloaded.  
- [ ] Previously when a tag that uses the `<source>` attribute was lazyloaded and the `preload` value was added there was also a `<noscript>` tag added which was redundant we need to check it is no longer added.

irinelenache:
@GrigoreMihai Tested and everything's fine now, thank you 🚀

PR #471, opened 2023-02-16
chore: fixed media offload page flag issues

GrigoreMihai:
Fixes https://github.com/Codeinwp/optimole-service/issues/673 . 

Fixes https://github.com/Codeinwp/optimole-service/issues/674 .

Changes:

Uses `POST_OFFLOADED_FLAG` for queries only if the plugin was installed after the flag was added: https://github.com/Codeinwp/optimole-wp/releases/tag/v3.5.0 .

Updates page meta with the `POST_OFFLOADED_FLAG` when an images is uploaded directly into the page. 

Removes redundant(wrong) `POST_OFFLOADED_FLAG` removal on page update.

### DIFF
--- a/inc/media_offload.php
+++ b/inc/media_offload.php
@@ -68,6 +68,12 @@ class Optml_Media_Offload extends Optml_App_Replacer {
 	 * @var string Used when moving the file to our servers.
 	 */
 	private static $last_deduplicated_original;
+	/**
+	 * Checks if the plugin was installed before adding POST_OFFLOADED_FLAG.
+	 *
+	 * @var bool Used when applying the flags for the page query.
+	 */
+	private static $is_legacy_install = null;
 
 	/**
 	 * Adds page meta query args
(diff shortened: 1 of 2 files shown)


inc/media_offload.php
```diff
@@ -68,6 +68,43 @@ class Optml_Media_Offload extends Optml_App_Replacer {
...
+		}
+		if ( $action === 'rollback_images' ) {
+			$args['meta_query'] = [
+				'relation' => 'AND',
+				[
+					'key' => self::POST_ROLLBACK_FLAG,
+					'compare' => 'NOT EXISTS',
+				],
+			];
+			if ( get_option( 'optimole_wp_install', 0 ) > 1666123200 ) {
```
selul, line 97: not super important but we can cache the option value inside a static var to avoid calling it each time. 

static $is_legacy_install = null; 
if($is_legacy_install === null ) { $is_legacy_install = get_option( 'optimole_wp_install', 0 ) > 1666123200 }

selul (approved):

irinelenache:
@GrigoreMihai Tested and both issues are fixed now 🚀

PR #477, opened 2023-02-23
Fix/449

### LINKED ISSUE #449: Coverting GIF to videos in some situations causes issues with other objects not loading
**Describe the bug*

A user reported that when he enables GIF to Video conversion in Optimole settings, the Contact form that is at the bottom of this page: https://lp.smarti.it/web-design/ stops loading.
This form: https://vertis.d.pr/i/Pm3h4g

After checking with Mihai he found out that we optimize the gif in this script https://vertis.d.pr/i/0nTITw

When the mentioned GIF is excluded it seems to work.
We should only optimize that gif, not lazyload and not convert to video.

The issue is discussed here - https://secure.helpscout.net/conversation/2026043174/339080

**WordPress Environment**
<details>
Site health information - https://pastebin.com/vaTsM2BC
</details>

GrigoreMihai:
Fixes #449 . 

Excludes gifs from the `plugins` directory and gifs with `min(width, height) <=20` from lazyload and video conversion. 

For testing this we can have some gifs added and set the width, height from the editor below the limit, they should be optimized and not be lazyloaded or converted to video when the option is enabled.

The same should happend for gifs from the `plugins` directory. For those we do not need to have an actual gif uploaded in the plugins folder it is enough to insert the URL without the image to exist (eg: `http://optimoletest.test/wp-content/plugins/test-plugin/something.gif` and just check that the tag is not lazyloaded/converted. 

The gifs under these conditions should only be optimized like this https://vertis.d.pr/i/rKoaNs .

### DIFF
--- a/inc/lazyload_replacer.php
+++ b/inc/lazyload_replacer.php
@@ -323,8 +323,6 @@ final class Optml_Lazyload_Replacer extends Optml_App_Replacer {
 		$search = [];
 		$replace = [];
 		foreach ( $video_tags as $video_tag ) {
-			do_action( 'optml_log', 'here' );
-			do_action( 'optml_log', $video_tag );
 
 			if ( ! $this->should_lazyload_iframe( $video_tag ) ) {
 				continue;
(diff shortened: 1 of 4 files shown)


GrigoreMihai:
The test is failing because of `CYPRESS_RECORD_KEY` secret , I think it expired @selul ?

selul (approved):

PR #478, opened 2023-02-23
Add build on deployment

### LINKED ISSUE #474: Generate build on deployment

Everytime you make changes, you have to add the build files to your PR and it's just useless step that we can avoid and generate it during the deployment.Every timea 

Like Neve or Otter, run the build during the deployment.

HardeepAsrani:
### All Submissions:

Changes to github actions workflow where we now generate the build file during the github action workflow instead of committing the files to git.

The e2e tests workflow's npm install but the task still didn't fail so we weren't aware of it as vue build files weren't required for the tests was failing so also fixed that.

Also VSCode formatted the other JS files, so no code change there, just prettify.

Closes #474.

1. Tested the QA & build-zip, can't really test the deployment but looking at the how other workflows work, it should work good too.
2. Not sure what else can be done to test here, tasked worked fine when tested with act.

### Other information:


.github/workflows/deploy.yml
```diff
@@ -12,11 +12,14 @@ jobs:
       - name: Install node version
         uses: actions/setup-node@v1
         with:
-          node-version: 12.x
-      - name: Build
+          node-version: 14.x
```
GrigoreMihai, line 15: It looks good, just a small mention. Let's use node version `14` for all workflows just in case there are any inconsistencies between 14 and 14.x .

GrigoreMihai (changes requested):

GrigoreMihai (approved):

PR #513, opened 2023-04-03
Fix Images with a Bullet symbol not being optimized

HardeepAsrani:
### All Submissions:

Previously, images with UTF character 2022, ie the bullet, weren't being optimized. The fix makes sure they are working properly.

Closes https://github.com/Codeinwp/optimole-service/issues/738

1. Upload an image with the bullet character in its filename, for example `DSCF4559-Edité-•-©-Josef-Helie.png`
2. Add it to your post, and confirm the Optimole CDN URL is being applied to it.

### Other information:


selul:
@HardeepAsrani can you add also an unit test, it's pretty straighforward to add one.

GrigoreMihai (approved):
LGTM, one small thing I would do is to also add an URL for this case for the default replacement here: https://github.com/Codeinwp/optimole-wp/blob/master/tests/test-replacer.php#L310 .

HardeepAsrani:
@GrigoreMihai Done.

rodica-andronache:
@HardeepAsrani hi! just wanted to make sure, this is supposed to fix the issue just with the bullet symbol ( which is does ).
Saw that for some other symbols, for e.g. `^` Optimole does not optimize the image (the final image here https://rodicaelena.com/moments/). Is that something I should create another issue for?

HardeepAsrani:
@GrigoreMihai Is there any reason why we don't just accept all the keyboard symbols with a wildcard? Could it cause any issues?

HardeepAsrani:
Thanks @GrigoreMihai, I've only added `^` caret support, seems like most of the keyboard characters are stripped out of the URL when you upload the image, so we don't need to support. Let me know if anything else is needed here.

Issue #505, opened 2023-03-31
Add Ability to Disable Noscript Tag in Optimole


selul:
### What problem does this address?

In order to provide greater flexibility for users, it would be beneficial to add the ability to disable the noscript tag in the Optimole plugin.
 

### What is your proposed solution?

In the lazyload tab, we can add a new toggle, called `Enable noscript tag` which by default will be off for new users, which means the no script tags will not load for them and on for users that have already Optimole connected. 

We can use this as a setting description: 

_The noscript tag offers fallback images for browsers that can't handle JavaScript-based lazy loading or related features. Disabling it may resolve conflicts with other plugins or configurations and decrease HTML page size._

Issue #520, opened 2023-04-11
Deprecated and Header Modification Errors When Offload is Enabled


kushh23:
### Description

A user reported that he is encountering issues with the Optimole plugin latest version when the offload feature is enabled on their WordPress site. The stack being used includes PHP 8.2 and WordPress 6.2

The following errors are being displayed:

> Deprecated: Creation of dynamic property Optml_Main::$media_offload is deprecated in /var/www/html/wp-content/plugins/optimole-wp/inc/main.php on line 82
> Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/optimole-wp/inc/main.php:82) in /var/www/html/wp-admin/includes/misc.php on line 1431
> Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/optimole-wp/inc/main.php:82) in /var/www/html/wp-includes/functions.php on line 6896
> Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/optimole-wp/inc/main.php:82) in /var/www/html/wp-includes/pluggable.php on line 1427
> Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/optimole-wp/inc/main.php:82) in /var/www/html/wp-includes/pluggable.php on line 1430

### Step-by-step reproduction instructions

1- Create an instance using these configurations
PHP version: 8.2
WordPress version: 6.2
Plugin version: Optimole 3.6.1

2- Check error logs.

### Screenshots, screen recording, code snippet or Help Scout ticket

https://secure.helpscout.net/conversation/2206769213/364850?folderId=2353196

### Environment info

https://pastebin.com/DaUGPQ0p

### Is the issue you are reporting a regression

No

kushh23:
+1: https://secure.helpscout.net/conversation/2212290317/365682?folderId=2353196

kushh23:
is [this ](https://secure.helpscout.net/conversation/2213419732/365857?folderId=2353196)the same case ? @GrigoreMihai

Issue #509, opened 2023-03-31
Enable lazyload for videos and iframes by default for new users.


selul:
### What problem does this address?

To improve the out-of-the-box experience and performance for new users, enable lazyload for videos and iframes by default in the Optimole plugin.

### What is your proposed solution?

1. Update the default settings in Optimole to have the lazyload feature for videos and iframes enabled upon installation.
2. Ensure that the updated settings do not affect existing users who have previously configured the plugin.

PR #521, opened 2023-04-12
Fixes

GrigoreMihai:
### All Submissions:

 For the iframe and video lazyload the option will be enabled for new users while for old install will stay the same. 

 Adds an option to control whether or not `<noscript>` tag should be added. For new install it will be off by default and for users that already had the plugin connected it will be on.

 Declares the `media_offload` property on the class to avoid depracation notice. 

Closes https://github.com/Codeinwp/optimole-wp/issues/505 ,  https://github.com/Codeinwp/optimole-wp/issues/509, https://github.com/Codeinwp/optimole-wp/issues/520.

For the noscript option, on the case of fresh installs. 

1. Install the pr plugin and connect it to an account.
2. Check that the noscript option is off by default.

For the noscript option, on the case when the plugin is already connected and the user updates. 

1. Install the current plugin version. 
2. Connect an account to it. 
3. Update the plugin with the pr version. 
4. Check that the plugin has the noscript option enabled. 
5. If you change the noscript option, lets say disable it, it should remain disabled.

For the iframe and video lazyload, on the case of fresh installs. 

1. Install the pr plugin and connect it to an account.
2. Check that the iframe and video lazyload is enabled. 

For the iframe option, on the case when the plugin is already connected and the user updates. 

1. Install the current plugin version. 
2. Connect an account to it. 
3. Update the plugin with the pr version. 
4. Check that the plugin has the iframe and video lazyload disabled as it was before updating the plugin. 

For the depracated notice. 

1. Make sure the test instance has `WP_DEBUG` on. 
2. Install the pr plugin. 
3. Check the debug logs.
4. The depracted notice from https://github.com/Codeinwp/optimole-wp/issues/520 should not appear.

### Other information:

* [x] Have you successfully ran tests with your changes locally?version.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -890,6 +890,8 @@ The root cause might be either a security plugin which blocks this feature or so
 				'enable_bg_lazyload_title'          => __( 'Enable lazyload for background images', 'optimole-wp' ),
 				'enable_video_lazyload_desc'        => __( 'Lazyload iframes/videos', 'optimole-wp' ),
 				'enable_video_lazyload_title'       => __( 'Enable lazyload for embedded videos and iframes.', 'optimole-wp' ),
+				'enable_noscript_desc'              => __( 'The noscript tag offers fallback images for browsers that can\'t handle JavaScript-based lazy loading or related features. Disabling it may resolve conflicts with other plugins or configurations and decrease HTML page size.', 'optimole-wp' ),
+				'enable_noscript_title'             => __( 'Enable noscript tag', 'optimole-wp' ),
 				'enable_gif_replace_title'          => __( 'Enable Gif to Video conversion', 'optimole-wp' ),
 				'enable_report_title'               => __( 'Enable error diagnosis tool', 'optimole-wp' ),
 				'enable_report_desc'                => __( 'Provides a troubleshooting mechanism which should help you identify any possible issues with your site using Optimole.', 'optimole-wp' ),
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -54,7 +54,8 @@ class Optml_Admin {
 	 */
 	public function init_no_script() {
 		if ( $this->settings->is_connected() ) {
-			if ( empty( $this->settings->get( 'no_script' ) ) ) {
+			$raw_settings = $this->settings->get_raw_settings();
+			if ( ! isset( $raw_settings['no_script'] ) ) {
 				$this->settings->update( 'no_script', 'enabled' );
 			}
 		}
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -38,6 +38,7 @@ class Optml_Admin {
 			add_action( 'init', [$this, 'check_domain_change'] );
 		}
 		add_action( 'admin_init', [ $this, 'maybe_redirect' ] );
+		add_action( 'admin_init', [ $this, 'init_no_script' ] );
 		if ( ! is_admin() && $this->settings->is_connected() && ! wp_next_scheduled( 'optml_daily_sync' ) ) {
 			wp_schedule_event( time() + 10, 'daily', 'optml_daily_sync', [] );
 		}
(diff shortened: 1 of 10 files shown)


inc/settings.php
```diff
@@ -105,6 +113,12 @@ class Optml_Settings {
 	 * Optml_Settings constructor.
 	 */
 	public function __construct() {
+
+		if ( self::$is_legacy_install === null ) {
+			// CHANGE THIS WHEN DEPLOYING LIVE
+			self::$is_legacy_install = get_option( 'optimole_wp_install', 0 ) < 1681313471;
```
selul, line 119: @GrigoreMihai I think there are better solutions than this since not all the users update the plugin at the same time. 

I would suggest just updating the value on some init hook or admin_init, or template_redirect based if the plugin is connected or not.

selul (changes requested):

inc/settings.php
GrigoreMihai: I think that the update time is not affecting this, because we want to know if the plugin was installed before or after the release to have the option enabled/disabled based on that. 

Not sure if you are referring that is not efficient to have this check here and instead we should update the `no_script` option only once on some hook ? (`admin_init` seems the best for this )

selul: No, I think this hack with time is  not something reliable and I would like to avoid it.

GrigoreMihai: Thank you for the clarification, I get it now.

GrigoreMihai: Changed it now if an user is updating the plugin and it is connected it will on by default , for a clean install it will be off.

inc/admin.php
```diff
@@ -48,6 +49,17 @@ public function __construct() {
 		}
 	}
 
+	/**
+	 * Init no_script setup value based on whether the user is connected or not.
+	 */
+	public function init_no_script() {
+		if ( $this->settings->is_connected() ) {
+			$raw_settings = $this->settings->get_raw_settings();
```
selul, line 57: @GrigoreMihai looking at this it seems that is always executing, no matter if the user is new or not. Considering that the user is new, once they connect the plugin, will go into this logic, no? 

Have you tested this logic?

GrigoreMihai: yup, for new users that are connecting the plugin it will be off, it my local testing is_connected is returning false at the begining and the default schema will be used before this check will go through , after the following calls will have the option setup from the default schema and it will be off unless the users changes this.

selul: @GrigoreMihai ok, I have looked over the code, and I don't see any place where we set up the option from the default schema; we only use it when getting the values, but we are not setting them. Since in your logic, you use raw values, the schema is ignored, at least, this is how I see it. 

Maybe I'm missing something, but I guess we can update the testing instructions to address this case.

selul:
@GrigoreMihai seems like the PR build logic is broken too.

inc/admin.php
GrigoreMihai: Yup it was intended to skip the schema to cover the case when the user updated the setting otherwise we get the default `disabled` and we have no way to know if the user set it or not. 

I updated the test comment. If it's ok with you I can move it to testing?

GrigoreMihai:
> @GrigoreMihai seems like the PR build logic is broken too as it adds multiple comments instead of updating it.

Will note it in an issue and look over it.

inc/admin.php
selul: @GrigoreMihai for me, the PR testing instructions look the same and not much helpful for the QA https://vertis.d.pr/i/dbWnH4

GrigoreMihai: For testing I used a separate comment here: https://github.com/Codeinwp/optimole-wp/pull/521#issuecomment-1513424105 , https://vertis.d.pr/i/YKhixR , the one you mentioned was only for detailing the code logic I thought it is better to not edit that , ofc I can also change that if do not think it's a good idea ?

selul: @GrigoreMihai, I noticed that the PR description could be more relevant to the context and should not contain outdated information. Since this is the third time I'm mentioning this, I kindly request you to follow the PR template closely and provide detailed testing instructions for the QA team in the description. You can refer to PRs in the Otter/Neve repositories as excellent examples. Thanks!

selul (approved):

rodica-andronache:
@GrigoreMihai everything's working well 👍

PR #523, opened 2023-04-18
added wildcard background lazy selector for wp_block_cover #758

GrigoreMihai:
Fixes https://github.com/Codeinwp/optimole-service/issues/758 . 

Using `[class*="wp-block-cover"][style*="background-image"]` will target both the old selector and the new ones that might be used.

### DIFF
--- a/inc/lazyload_replacer.php
+++ b/inc/lazyload_replacer.php
@@ -124,7 +124,7 @@ final class Optml_Lazyload_Replacer extends Optml_App_Replacer {
 		$default_watchers = [
 			'.elementor-section[data-settings*="background_background"]',
 			'.elementor-section > .elementor-background-overlay',
-			'.wp-block-cover[style*="background-image"]',
+			'[class*="wp-block-cover"][style*="background-image"]',
 		];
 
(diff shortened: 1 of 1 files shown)


GrigoreMihai:
For testing this we need to check that the cover block has the background lazyloaded when it is setup with the background image option: https://vertis.d.pr/i/Psiohb , https://vertis.d.pr/i/uxD9q1 .

rodica-andronache:
hi @GrigoreMihai `optml-bg-lazyloaded` is added only if `Fixed Background` or `Repeated Background` are enabled. Is that the expected behaviour? https://vertis.d.pr/i/jpi00d

GrigoreMihai:
> hi @GrigoreMihai `optml-bg-lazyloaded` is added only if `Fixed Background` or `Repeated Background` are enabled. Is that the expected behaviour? https://vertis.d.pr/i/jpi00d

Hello @rodica-andronache it is the expected behaviour without using those setups the block is not actually using the image as a background image, it is setup using `src` https://vertis.d.pr/i/msJYnp .

PR #538, opened 2023-05-12
feat: implement maxWidth and maxHeight setting

### LINKED ISSUE #506: Max Width or Height for Image Resizing in Optimole

Introduce a feature in Optimole that allows users to define a max width or height for images, which will be used when Optimole resizes images. If an image in the content is displayed at a size larger than the defined max dimensions, the image will be resized to the specified max width or height.

By providing this feature, users can gain more control over their image dimensions, ensuring consistent image sizes across their websites and potentially improving overall website performance.

1. Add an option in Optimole settings to enable this feature using a toggle inside the Resize tab and allow users to input their desired max width or height.
2. Implement the resizing logic in Optimole to ensure that images larger than the specified max dimensions are automatically resized to the user-defined max width or height.

This will be by default off for old users, who already have Optimole connected, and for new users, this will be enabled they can use 1920 for width and 1080 for height. 

The section can use this title: `Limit Image Dimensions with Max Width/Height`
The setting can use this description: 
_This feature allows you to set a maximum width or height for images on your website, automatically resizing larger images to fit within the defined limits while maintaining the original aspect ratio._

When this is enabled, we can disable the retina part to make it efficient. We can also add a note there like: 

_When you enable this feature to define a max width or height for image resizing, please note that DPR (retina) images will be disabled. This is done to ensure consistency in image dimensions across your website. Although this may result in slightly lower image quality for high-resolution displays, it will help maintain uniform image sizes, improving your website's overall layout and potentially boosting performance._ 

How this will be applied: 

- In the Javascript library that does the resize, this will be defined on the script init arguments, as maxWidth and maxHeight. When they are not defined, the behavior considering the ratio will be like this. 
If an image has a requested size of 3000x2000 pixels and the user-defined max width is 1920 pixels and max height is 1080 pixels, the scaling factors for width and height would be 1920/3000 and 1080/2000, respectively. The smaller scaling factor (1080/2000) is chosen, and both dimensions are multiplied by this factor to obtain the new dimensions: 1620x1080 pixels.

- On the PHP end, if the width/height is not auto, we can apply the same logic as above ☝️ otherwise, we will use instead of width/height auto these new values, without the enlarging option, in the way the engine won't enlarge the images if they are smaller but it will resize them if they are bigger

abaicus:
### All Submissions:

- Adds a new option to limit max-width and max-height of the served images;
- This option will be enabled by default for new users;
- This will limit the image sizes served on the frontend in the following way:
  - The scale ratio is established by applying the smallest ratio between maxWidth / initialWidth and maxHeight / initialHeight;
  - Then, the image initial w/h is multiplied by this ratio, obtaining the new size for generating the URL w:/h: parameters;
- Images that were served with `w`/`h` auto will be automatically be served with the **w:maxWidth** and **h:maxHeight**

Closes #506.

1. This should be tested using a staging API key. The version of the JS library that supports this should already be available (https://github.com/Codeinwp/optimole_js_lib/pull/123); 
2. Make sure that the option is enabled under the **Optimole Dashboard > Settings > Resize > Limit Image Dimensions with max width/height** (this option will be enabled by default on an instance where optimole was never installed before today);
3. Make sure you have images that exceed the defined size (1920⨉1080 by default), but you can try a much more visible setup that impacts the image size much more visible - something like 300x200 or similar;
4. The requested URLs for these images on the front end should not exceed the width or the height defined there as long as the option is enabled. This should happen to both the images loaded initially and the lazyloaded ones; 

### Other information:

### DIFF
--- a/inc/app_replacer.php
+++ b/inc/app_replacer.php
@@ -69,6 +69,24 @@ abstract class Optml_App_Replacer {
 	 */
 	protected $max_height = 3000;
 	/**
+	 * Defines if the dimensions should be limited when images are served.
+	 *
+	 * @var int
+	 */
+	protected $limit_dimensions_enabled = false;
+	/**
+	 * Defines which is the maximum width accepted when images are served.
+	 *
+	 * @var int
+	 */
+	protected $limit_width = 1920;
+	/**
+	 * Defines which is the maximum height accepted when images are served.
+	 *
+	 * @var int
+	 */
+	protected $limit_height = 1080;
+	/**
 	 * Defines if css minification should be used.
 	 *
 	 * @var int
@@ -141,6 +159,8 @@ abstract class Optml_App_Replacer {
 	 * @var string Cache Buster value.
 	 */
 	protected $active_cache_buster_assets = '';
+
+	const DIRECT_UPLOAD_SOURCE = 'directUpload';
 
 	/**
 	 * Returns possible src attributes.
@@ -404,7 +424,9 @@ abstract class Optml_App_Replacer {
 
 		$this->possible_sources = $this->extract_domain_from_urls(
 			array_merge(
-				[ get_home_url() ],
+				[
+					get_home_url(),
+				],
 				array_values( $this->site_mappings ),
 				array_keys( $this->site_mappings )
 			)
@@ -429,6 +451,12 @@ abstract class Optml_App_Replacer {
 
 		$this->max_height = $this->settings->get( 'max_height' );
 		$this->max_width  = $this->settings->get( 'max_width' );
+
+		$this->limit_dimensions_enabled = $this->settings->get( 'limit_dimensions' ) === 'enabled';
+		if ( $this->limit_dimensions_enabled ) {
+			$this->limit_height = $this->settings->get( 'limit_height' );
+			$this->limit_width  = $this->settings->get( 'limit_width' );
+		}
 
 		$this->is_css_minify_on = ( $this->settings->get( 'css_minify' ) === 'enabled' ) ? 1 : 0;
 		$this->is_js_minify_on  = ( $this->settings->get( 'js_minify' ) === 'enabled' ) ? 1 : 0;
--- a/inc/app_replacer.php
+++ b/inc/app_replacer.php
@@ -160,8 +160,6 @@ abstract class Optml_App_Replacer {
 	 */
 	protected $active_cache_buster_assets = '';
 
-	const DIRECT_UPLOAD_SOURCE = 'directUpload';
-
 	/**
 	 * Returns possible src attributes.
 	 *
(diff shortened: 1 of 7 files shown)


inc/app_replacer.php
```diff
@@ -142,6 +160,8 @@ abstract class Optml_App_Replacer {
 	 */
 	protected $active_cache_buster_assets = '';
 
+	const DIRECT_UPLOAD_SOURCE = 'directUpload';
```
GrigoreMihai, line 163: This seems redundant as is not used anywhere. I think it is safe to remove if I'm not missing something.

GrigoreMihai (approved):
Looks good, just the small thing I mentioned.

inc/app_replacer.php
abaicus: @GrigoreMihai, Thanks for mentioning this! Indeed, it's a leftover from testing! 👍🏻 I'll remove it.

irinelenache:
@abaicus Nice job 🚀 Tested and found a single problem, let me know if it's expected or not:

- [ ] Using the staging API key without adding this PR build, i'm getting a warning next to an image in the shop page https://vertis.d.pr/i/EMleGO . After adding this PR build, the warning is gone

You can check on this instance:
```
https://ng-of-doom-fugi.vertisite.cloud/wp-admin
wahozimava5946
nlST6AmJaybNYLrKcfE9
```

abaicus:
@irinelenache Thanks for the heads-up! 🚀 

It should have nothing to do with this PR. The staging API would only use the updated JS library from this PR: https://github.com/Codeinwp/optimole_js_lib/pull/123, while the warning seems to be on the PHP side. Also, the file comes from the wp-admin folder. 

Specifically, the error stems from here: https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/image.php#L845-L852

If the `WP_DEBUG` constant is not defined, then the `exif_read_data` function is running with the error suppression flag (`@`).

Seems like there's an ongoing core issue that hasn't been touched for a few years: https://core.trac.wordpress.org/ticket/42480


All in all, I think this has nothing to do with this specific PR. 👍🏻

irinelenache:
@abaicus Thanks for explaining the issue, i'll move this to **Ready to merge** then 🚀

inc/main.php
```diff
@@ -144,6 +145,26 @@ public static function change_lazyload_default( $defaults ) {
...
+	 *
+	 * @param array $defaults Old defaults.
+	 *
+	 * @return array New defaults.
+	 */
+	public static function change_limits_default( $defaults ) {
+		$install_time = get_option( 'optimole_wp_install', 0 );
+
+		// Todo: change this before release to release day.
+		if ( $install_time < 1683870956 ) {
```
selul, line 159: @abaicus I would use the timestamp here since not all the people update in the same time and is not that reliable. We can use if the user is connected or not, if is conected we consider old, if he is not connected we consider new. @GrigoreMihai you used a similar approach on noscript approach

selul (changes requested):

inc/main.php
abaicus: @selul Thanks for letting me know. I changed the way we treat the setting based on the connection status.

selul (approved):

PR #553, opened 2023-05-26
Development

abaicus:
- add option to let users set max-width and max-height at which images are served


irinelenache:
@abaicus Had a final check and everything's fine regarding this release 🚀

abaicus:
Thank you very much for the help, @irinelenache 🚀

PR #630, opened 2023-08-07
Release

abaicus:
### Bug Fixes
- **Fatal error on PHP<7.3**: Fixed fatal error that the plugin was throwing on PHP versions lower than 7.3


irinelenache:
@abaicus Tested with PHP 5.6 - 8.2 and didn't encounter any error, sorry for missing this before the release 🙏

Issue #633, opened 2023-08-09
lazyload and resize are by default off instead of ON


selul:
### Description

In the most recent versions, since we updated the dashboard, the default value for the lazyload feature is OFF vs ON as it should be.

This has a major impact for users that are trying Optimole for the first time since they need to enable that manually. 

We should have proper unit tests to avoid this from happening in the future. 

### Step-by-step reproduction instructions

1. Create a new WP env and Connect Optimole to latest version and check Resize and Lazyload option, is showing as OFF
2. Create a new WP env Connect Optimole on the 3.7 version and check the Resize and Lazyload option
 

### Screenshots, screen recording, code snippet or Help Scout ticket

https://vertis.d.pr/i/Fa14Jl

### Is the issue you are reporting a regression

Yes, this is a regression.

Issue #632, opened 2023-08-09
fatal error on preloader with Elementor


selul:
### Description

It seems like some users are reporting issues with the preloader functionality in Optimole + Elementor.

### Step-by-step reproduction instructions

Details here https://secure.helpscout.net/conversation/2323544414/380749?folderId=2353201

### Is the issue you are reporting a regression

Yes, this is a regression.

PR #635, opened 2023-08-09
Release

abaicus:
### Bug Fixes
- **Fatal error**: In some contexts, the Hero Preloader feature was throwing a fatal error, trying to access inexistent posts.
- **Lazyload default status**: Lazyload was off for new users by default when it should have been on.

--- 

### Testing instructions:

For #633: 
- Create a new instance and install the plugin;
- Activate the plugin and connect to the service;
- The setting Scale and Lazyload images should be ON by default;

For #632:
- I wasn't able to replicate this issue. I will try to confirm with the user that the fix works fine;

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -104,6 +104,7 @@ class Optml_Admin {
 		}
 
 		$this->settings->update( 'limit_dimensions', 'enabled' );
+		$this->settings->update( 'lazyload', 'enabled' );
 
 		update_option( self::NEW_USER_DEFAULTS_UPDATED, 'yes' );
 	}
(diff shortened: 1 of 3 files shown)


irinelenache:
@abaicus Tested and the lazyload issue is fixed 🚀 I also tried and couldn't replicate #632 , if you find a way let me know and i'll test again 👍

abaicus:
@irinelenache thank you for testing this! 🙏🏻 

I tried reaching out to the client and testing this on his environment, but he didn't respond yet. I will deploy this as it is, and hopefully it should fix his specific issue.

PR #658, opened 2023-08-30
Feat/improve woo lazyload

GrigoreMihai:
### All Submissions:

Previously when we were lazyloading the images in the zoom slider we were resizing the image picked there which should have been in full size to avoid a blurry view. Currently this is not the case anymore because woocomerce changed the way they pick that image and are now taking it from a `data-src` attribute: https://vertis.d.pr/i/uU7CFv . 

The other issue was that the transition for the lazyloaded images was not smooth, the blur effect (or a white space in the placeholder case) where visible when transitioning to images that needed to be lazyloaded. 

This is fixed by early loading the images in the slider after everything else is loaded so it does not affect the page initial load time. For this part we also need to deploy: https://github.com/Codeinwp/optimole_js_lib/pull/128. 

This how the blur would have looked beforet the change: https://vertis.d.pr/v/Bfs7oK and with the change that is dissapearing: https://vertis.d.pr/v/VmntBD .  

The thumbnails that are used to swap the images in slider cannot be lazyload because they are generated using javascript. Also they are always in the viewport so there is no gain in lazyloading those.

The first image in the slider (the default selected) will not be lazyloaded because of our exclude first images option.

Part of https://github.com/Codeinwp/optimole-service/issues/760 .

Small notes from the research: 

- The other lazyload plugins are not lazyloading the slider images, I think this is mainly because they are using `data-src` there. 
- The slider plugins are using their custom approach, but none of the ones I tested used the `data_large-image` class, therefore the sliders should not be affect by this change

1. Install the pr plugin.
2. Setup woocomerce and a few products that have a products gallery containing multiple images.
3. Check that the product gallery works as expected with the lazyload enabled, main things to look for: the slider transitions and the zoom for the selected image works as expected. 
4. Install and test different woocommerce gallery plugins to make sure this change is not breaking any of them. The ones I found to be the most used are these: https://vertis.d.pr/i/ZRtUZa , those are just a suggestion feel free to try others if you wish. 

### Other information:

### DIFF
--- a/inc/lazyload_replacer.php
+++ b/inc/lazyload_replacer.php
@@ -220,7 +220,7 @@ final class Optml_Lazyload_Replacer extends Optml_App_Replacer {
 	/**
 	 * Check if the img tag should be lazyloaded early.
 	 *
-	 * @param $image_tag string The image tag.
+	 * @param string $image_tag The image tag.
 	 * @return bool Whether the image tag should be lazyloaded early or not.
 	 */
 	public function should_lazyload_early( $image_tag ) {


abaicus (approved):

irinelenache:
@GrigoreMihai Tested and here is what i found:

- [ ] After a refresh, the woocommerce gallery images look blurry unless they are zoomed by hovering over them. If the window is resized (i noticed this after opening and closing the devtools) the images look as they should. Here is a video with the behaviour https://vertis.d.pr/v/4zcUr1
- [ ] I didn't exactly understand if the gallery images should/shouldn't be lazyloaded, because for now they are not https://vertis.d.pr/i/mNu6fs

_Note: Testing is not complete on this PR_

GrigoreMihai:
@irinelenache The first issue is fixed now, be sure to clear the browser cache before testing again. 

For the second question, the gallery images should not be lazyloaded with a default setup (due to the impossibility to do it).

Some gallery plugins are having a different flow that is allowing us to lazyload those and in those cases we do lazyload the gallery.  

To summarize there is no issue if those are lazyload/not lazyloaded depending on the setup.

irinelenache:
@GrigoreMihai The issue is fixed now, thank you 👍 I also tested the compatibilities with the gallery plugins and didn't find any problem 🚀

PR #697, opened 2023-09-28
feat: remove Resize large images original source

### LINKED ISSUE #675: Remove Resize large images original source.

Right now, we have two features that somehow share the functionalities among them up to a point. For the sake of simplicity, I will call them: 

- **RLS** - Resize large images from the original source.
- **LMD** - Limit Image Dimensions with max width/height

On the PHP Side, RLS is doing the resize to those boundaries if the image size is [known](https://github.com/Codeinwp/optimole-wp/blob/master/inc/url_replacer.php#L254-L260) ( not auto ) while the LMD is [covering](https://github.com/Codeinwp/optimole-wp/blob/master/inc/url_replacer.php#L229-L234) the auto case as well, by forcing those boundaries by default to the auto images as well. 

Moreover, the LMD is also covering the Autoscalling feature, which the RLS is not, since is doing only the backend stuff, which is pre-browser resizing or lazyload. 

I suggest remove RLS together, don't merge any value, and keep the LMD as it is. 

The impact of this is that old users who were having RLS will no longer use it, and to accommodate this; I would suggest we turn on LMD as well for legacy users. 

@Codeinwp/optimole do you see any case that could not be covered by this? Like a use case or similar that would be affected by the change. 

Keeping you in the loop @Codeinwp/design-team 

### Will this feature require documentation? (Optional)

Yes, it requires documentation.

abaicus:
### All Submissions:

- Removes the `Resize large images original source` setting;
- Enables the `Limit Image Dimensions` setting for old users;
- Changes the `Limit Image Dimensions` setting to be enabled by default in all cases; 

Closes #675.

1. Go to the Optimole Settings page (**Optimole Dashboard > Settings > Resize**) 
2. The `Resize large images original source` setting should not exist anymore;
3. For users that updated the plugin, the `Limit Image Dimensions` setting should be turned on now;
4. For new users, the `Limit Image Dimensions` setting should be turned on by default;
5. The setting should function just as before (https://github.com/Codeinwp/optimole-wp/pull/538), resizing images larger than 1920x1080 (or whatever limit is set under the `Limit Image Dimensions` setting);
6. You can check this by looking at the URL in the DOM - w/h should not exceed 1920/1080;

### Other information:

* ~[] Have you written new tests for your changes, as applicable?~

### DIFF
--- a/inc/url_replacer.php
+++ b/inc/url_replacer.php
@@ -251,16 +251,10 @@ final class Optml_Url_Replacer extends Optml_App_Replacer {
 			}
 		}
 
-		if ( $args['width'] > 0 && $args['height'] > 0 ) {
-			list( $args['width'], $args['height'] ) = wp_constrain_dimensions( $args['width'], $args['height'], $this->max_width, $this->max_height );
-		} elseif ( $args['width'] > 0 ) {
-			$args['width'] = $args['width'] > $this->max_width ? $this->max_width : $args['width'];
-		} elseif ( $args['height'] > 0 ) {
-			$args['height'] = $args['height'] > $this->max_height ? $this->max_height : $args['height'];
-		}
 		if ( isset( $args['resize'], $args['resize']['gravity'] ) && $this->settings->is_smart_cropping() ) {
 			$args['resize']['gravity'] = Optml_Resize::GRAVITY_SMART;
 		}
+
 		$args = apply_filters( 'optml_image_args', $args, $original_url );
 
 		$arguments = [
--- a/assets/src/dashboard/parts/connected/settings/Resize.js
+++ b/assets/src/dashboard/parts/connected/settings/Resize.js
@@ -239,37 +239,6 @@ const Resize = ({
 					) ) }
 				</BaseControl>
 			) }
-
-			<hr className="my-8 border-grayish-blue"/>
-
-			<BaseControl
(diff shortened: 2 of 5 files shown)


inc/url_replacer.php
```diff
@@ -251,16 +251,6 @@ private function normalize_image( $url, $original_url, $args, $is_uploaded = fal
...
 		}
 
-		if ( $args['width'] > 0 && $args['height'] > 0 ) {
-			list( $args['width'], $args['height'] ) = wp_constrain_dimensions( $args['width'], $args['height'], $this->max_width, $this->max_height );
-		} elseif ( $args['width'] > 0 ) {
-			$args['width'] = $args['width'] > $this->max_width ? $this->max_width : $args['width'];
-		} elseif ( $args['height'] > 0 ) {
-			$args['height'] = $args['height'] > $this->max_height ? $this->max_height : $args['height'];
-		}
-		if ( isset( $args['resize'], $args['resize']['gravity'] ) && $this->settings->is_smart_cropping() ) {
```
selul, line 261: @abaicus this shouldnt be removed since is related to smart crop, no?

abaicus: @selul yes, indeed. Sorry for that! Thank you for letting me know!

GrigoreMihai (approved):

rodica-andronache:
@abaicus tested and everything's working well 🚀

PR #794, opened 2024-04-18
Development

selul:
- Addressed an issue with offloading/rollback transfers getting stuck by adding a check for scheduled cron jobs. If no cron is scheduled and a transfer is in progress, the cron will be rescheduled.


abaicus (approved):
Looks good! 🚀

PR #858, opened 2024-11-25
make `--type` option optional with `clear_cache` command

carlalexander:
Small fix to the `clear_cache` command that I found fixing an issue for a client. It says `--type` defaults to images, but that doesn't work in practice.

image


carlalexander:
One quick question is should we change `type` to be `images` instead of `''`. Otherwise, the success message doesn't work.

selul:
oh, you are right, thank you @carlalexander  🚀

PR #886, opened 2025-03-13
feat: improve plugin first screen

girishpanchal30:
### All Submissions:

Improve the account connections page layout as suggested [here](https://github.com/Codeinwp/optimole-service/issues/1335#issuecomment-2687267604).

image

Closes https://github.com/Codeinwp/optimole-service/issues/1335

### Other information:

### DIFF
--- a/assets/src/dashboard/parts/connect/index.js
+++ b/assets/src/dashboard/parts/connect/index.js
@@ -168,7 +168,7 @@ const ConnectLayout = () => {
 				<div className="optml-connect__content basis-8/12">
 					<div className="inline-block bg-light-blue text-gray-500 text-sm font-medium px-4 py-1.5 rounded-full mb-4">{ optimoleDashboardApp.strings.account_needed_trust_badge }</div>
 
-					<div className="text-gray-700 font-serif text-26 font-bold leading-7 m-0">{ optimoleDashboardApp.strings.account_needed_heading }</div>
+					<div className="text-gray-700 font-serif text-2xl font-bold leading-7 m-0">{ optimoleDashboardApp.strings.account_needed_heading }</div>
 					<p
 						className="text-base font-normal text-gray-500"
 						dangerouslySetInnerHTML={ { __html: optimoleDashboardApp.strings.account_needed_sub_heading } }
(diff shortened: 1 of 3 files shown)


abaicus (changes requested):
@girishpanchal30 let's refrain from using classes with dynamically defined values like `bg-[#f0f6fc]`, or `text-[#646970]`, `text-[#2271b1]`, etc. 

We can surely either match the colors to some variation of the custom-defined colors, or predefined tailwind values

girishpanchal30:
@abaicus Thank you for the review! I've fixed the feedback. Please review and let me know if any further changes are needed.

assets/src/dashboard/parts/connect/index.js
```diff
@@ -187,6 +199,14 @@ const ConnectLayout = () => {
 							dangerouslySetInnerHTML={ { __html: optimoleDashboardApp.strings.account_needed_subtitle_2 } }
 						/>
 					</div>
+
+					<div className="flex py-3">
```
abaicus, line 203: We can also add a `items-center` class on all three items in this list so that the checkmark is centered. Maybe also map them to an array with the contents we can loop through and output them so we don't have duplicated markup.

abaicus (changes requested):
@girishpanchal30 I meant using variables that already exist in the config, instead of introducing new ones either inline, or in the config itself. We can ballpark the design, and don't need it to be pixel-perfect to the artifact

- `slate-gray` can be matched to `gray-500`;
- The value of `text-s` is actually the font-size of the in the admin, so we could simply not add it;
- `text-15`, we can simply use `text-sm` which is 1px lower;

assets/src/dashboard/parts/connect/index.js
girishpanchal30: @abaicus fixed all suggestions

abaicus:
@girishpanchal30 I see there are still some leftover variables lef tin tailwind config. Can't we replace text-26 with text-2xl or something to that affect?

`text-s` is also leftover.

girishpanchal30:
> @girishpanchal30 I see there are still some leftover variables lef tin tailwind config. Can't we replace text-26 with text-2xl or something to that affect?

Removed text-26 class

> text-s is also leftover

I’ve already removed it from where it was previously used. Could you please recheck with the latest commit?

abaicus:
Thanks @girishpanchal30 ! Looks good! 👍🏻

abaicus (approved):

PR #898, opened 2025-03-21
Add optimizaiton tips in sidebar

girishpanchal30:
### All Submissions:

Add new sidebar box.
 

Closes https://github.com/Codeinwp/optimole-service/issues/1347

### Other information:

### DIFF
--- a/assets/src/dashboard/parts/connected/OptimizationStatus.js
+++ b/assets/src/dashboard/parts/connected/OptimizationStatus.js
@@ -1,0 +1,77 @@
+/**
+ * WordPress dependencies.
+ */
+import { Icon } from '@wordpress/components';
+import { closeSmall, check } from '@wordpress/icons';
+
+import { useSelect } from '@wordpress/data';
+
+const OptimizationStatus = () => {
+	const {
+		statuses
+	} = useSelect( select => {
+		const { getSiteSettings } = select( 'optimole' );
+
+		const siteSettings = getSiteSettings();
+
+		const statuses = [
+			{
+				active: 'enabled' === siteSettings?.image_replacer,
+				label: optimoleDashboardApp.strings.optimization_status.statusTitle1,
+				description: optimoleDashboardApp.strings.optimization_status.statusSubTitle1
+			},
+			{
+				active: 'enabled' === siteSettings?.lazyload,
+				label: optimoleDashboardApp.strings.optimization_status.statusTitle2,
+				description: optimoleDashboardApp.strings.optimization_status.statusSubTitle2
+			},
+			{
+				active: 'enabled' === siteSettings?.scale,
+				label: optimoleDashboardApp.strings.optimization_status.statusTitle3,
+				description: optimoleDashboardApp.strings.optimization_status.statusSubTitle3
+			}
+		];
+
+		return {
+			statuses: statuses
+		};
+	});
+
+	return (
+		<div className="bg-white flex flex-col text-gray-700 border-0 rounded-lg shadow-md p-8">
+			<h3 className="text-base m-0">{ optimoleDashboardApp.strings.optimization_status.title }</h3>
+			<ul>
+				{ statuses.map( ( status, index ) => {
+					let statusClass = status.active ? 'success' : 'danger';
+					return (
+						<li
+							key={ index }
+							className="flex items-start gap-2"
+						>
+							{ status.active ? (
+								<Icon icon={check} className="fill-success bg-success/20 rounded-full" size={20} />
+							) : (
+								<Icon icon={closeSmall} className="fill-danger bg-danger/20 rounded-full" size={20} />
+							) }
+
+							<div>
+								<span className='text-gray-700 font-normal font-semibold'>
+									{ status.label }
+								</span>
+								<p className="m-0">{ status.description }</p>
+							</div>
+						</li>
+					);
+				}) }
+			</ul>
+			<p
+				className="m-0 mt-3"
+				dangerouslySetInnerHTML={ {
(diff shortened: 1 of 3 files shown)


abaicus (changes requested):
@girishpanchal30 I left some comments on the issue. Let's keep consistent sizing between the SPC recommendation list / heading, and this new box in the sidebar.

Other than that I've also asked @vytisbulkevicius for some input on how the checkboxes should behave.

Thank you!

girishpanchal30:
@abaicus Any idea why `fill-danger bg-danger/20` isn't working on the SVG?

girishpanchal30:
@abaicus I've fixed the all feedbacks and now it look like below.

image

assets/src/dashboard/parts/connected/OptimizationStatus.js
```diff
@@ -0,0 +1,77 @@
...
+				active: 'enabled' === siteSettings?.scale,
+				label: optimoleDashboardApp.strings.optimization_status.statusTitle3,
+				description: optimoleDashboardApp.strings.optimization_status.statusSubTitle3
+			}
+		];
+
+		return {
+			statuses: statuses
+		};
+	});
```
abaicus, line 38: > [!IMPORTANT]
> It seems that the `scale` settings is somehow inverted. So we have to check if it's `disabled` instead of `enabled`. 
> I don't recall exactly what happened here, but it's probably some backwards compatibility.

--- 

@vytisbulkevicius Here when we toggle the respective settings, the sidebar widget is **not** updated unless you save the settings. I think this might be confusing to users.

Also I think having them as red, without any indication what setting the user should change to make them all green might be frustrating and require the user to dive into the settings.

Do you think we should have them real-time updated? 

@girishpanchal30 if @vytisbulkevicius agrees that they should be toggled with the settings value, instead of on save, we should pass the setting props through the component tree from [here](https://github.com/Codeinwp/optimole-wp/blob/eaae2096c4cf7123752e8683a9454a5313d26f71/assets/src/dashboard/parts/connected/index.js#L98), and make sure we keep things in sync as they are updated. 

Otherwise, there is no need for destructuring and returning an object here:

```suggestion
	const statuses = useSelect( select => {
		const { getSiteSettings } = select( 'optimole' );

		const siteSettings = getSiteSettings();

		return [
			{
				active: 'enabled' === siteSettings?.image_replacer,
				label: optimoleDashboardApp.strings.optimization_status.statusTitle1,
				description: optimoleDashboardApp.strings.optimization_status.statusSubTitle1
			},
			{
				active: 'enabled' === siteSettings?.lazyload,
				label: optimoleDashboardApp.strings.optimization_status.statusTitle2,
				description: optimoleDashboardApp.strings.optimization_status.statusSubTitle2
			},
			{
			    // 👇🏻 We should make sure to change this 👇🏻 
				active: 'disabled' === siteSettings?.scale,
				label: optimoleDashboardApp.strings.optimization_status.statusTitle3,
				description: optimoleDashboardApp.strings.optimization_status.statusSubTitle3
			}
		];
	});
```

abaicus (commented):
@girishpanchal30 besides what I mentioned, things look good. 

Let's wait for @vytisbulkevicius's input and see how we further update this. 

Thank you! 👍🏻

assets/src/dashboard/parts/connected/OptimizationStatus.js
girishpanchal30: I noticed that `'disabled' !==` is used in other parts of the code as well -  https://github.com/search?q=repo%3ACodeinwp%2Foptimole-wp%20%27disabled%27%20!%3D%3D&type=code

abaicus: For `scale` it is always checked the other way around: 

https://github.com/Codeinwp/optimole-wp/blob/eaae2096c4cf7123752e8683a9454a5313d26f71/assets/src/dashboard/parts/connected/settings/Lazyload.js#L39

https://github.com/Codeinwp/optimole-wp/blob/eaae2096c4cf7123752e8683a9454a5313d26f71/inc/admin.php#L562

abaicus: Does the optimization stat for `scaling` stay as green when the toggle is set to enabled for you and the setting is set to enabled with the current version on this branch?

vytisbulkevicius: @abaicus, @girishpanchal30 

> Here when we toggle the respective settings, the sidebar widget is not updated unless you save the settings. I think this might be confusing to users.

I don't think this is a problem. For a feature to work you need to save it so only enabling and changing the status could be even more confusing that you already enabled it and don't need to save the settings anymore. So IMO it's fine and we can keep it this way.

girishpanchal30: @abaicus Thanks! I've updated the condition in the latest commit.

girishpanchal30: @abaicus Should I mark this as resolved?

abaicus: Can you also update the destructuring please, like i mentioned in the initial comment?

girishpanchal30: @abaicus Done as suggested!

abaicus: Great! Thank you @girishpanchal30! 👍🏻

abaicus (approved):

PR #930, opened 2025-05-01
Style: Update dashboard colors and dependencies

ineagu:
the original issue didn't had a accurate mockup, using the style of the plugin and the expected improvements, ended up not being present. 

this commit aims to pick a lighter color for the bg + a lighter secondary button color since the "i already have an account" would be used in a very very few cases.

Screenshot 2025-05-01 at 10 48 23
Screenshot 2025-05-01 at 10 48 20

### DIFF
--- a/assets/src/dashboard/style.scss
+++ b/assets/src/dashboard/style.scss
@@ -6,8 +6,9 @@ $primary: #EF686B;
 $success: #5F9D61;
 $danger: #E77777;
 $warning: #ffebeb;
+$purple: #8b5cf6;
 $info: #577BF9;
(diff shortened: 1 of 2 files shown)


ineagu:
@abaicus when you have a chance would help to let me know if I approached this in the right way, unfortunately, the build is failing so I couldn't test with the PR built, did a build locally and tested it worked, but I updated also the tailwind config.

abaicus:
@ineagu yes, the approach looks fine 👍🏻 

We should check that other buttons don't look bad though.

PS: Build is failing because you're doing a PR from the personal fork. 😄

PR #960, opened 2025-09-16
Feat/improved scaling

selul:
https://github.com/Codeinwp/optimole-service/issues/1524

### DIFF
--- a/inc/compatibilities/kinsta.php
+++ b/inc/compatibilities/kinsta.php
@@ -1,0 +1,54 @@
+<?php
+
+/**
+ * Class Optml_kinsta.
+ *
+ * @reason Clear cache on kinsta.
+ */
+class Optml_kinsta extends Optml_compatibility {
+
+
+	/**
+	 * Should we load the integration logic.
+	 *
+	 * @return bool Should we load.
+	 */
+	public function should_load() {
+
+		return class_exists( '\CDN_Clear_Cache_Api' );
+	}
+
+	/**
+	 * Register integration details.
+	 *
+	 * @return void
+	 */
+	public function register() {
+		add_action( 'optml_clear_cache', [ $this, 'add_clear_cache_action' ], 10, 1 );
+	}
+
+
+	/**
+	 * Should we early load the compatibility?
+	 *
+	 * @return bool Whether to load the compatibility or not.
+	 */
+	public function should_load_early() {
+		return true;
+	}
+	/**
+	 * Clear cache on kinsta.
+	 *
+	 * @param string|bool $location The location to clear the cache for. If true, clear the cache globally. If a string, clear the cache for a particular url.
+	 * @return void
+	 */
+	public function add_clear_cache_action( $location ) {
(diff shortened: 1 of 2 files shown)


ineagu:
1. I couldn't get the caching clearance to work as expected on wpshout @selul . I am not sure is because of Kinsta cache, this works if I clear wp-rocket cache, which I think trigger Kinsta clearing as well. 

2. The solution that we are using to fix the sizes for the images that were previously lazy-loaded doesn't seem to work and I am not sure if we can find a solution : 

This image is originally loaded like this on desktop : Fastest WordPress Hosting With No Caching

After profiling, is being loaded like this : Fastest WordPress Hosting With No Caching   


It got the width height from the mobile profiling : Fastest WordPress Hosting With No Caching.

ineagu:
> @ineagu please share the page.

https://wpshout.com

ineagu:
Another test I did : 

https://testing-scaling.vertisite.cloud/elementor-101/

- we correctly identify LCP on both mobile/desktop ( it is a bg image ),  same as reported by Google pagespeed
Screenshot 2025-10-08 at 10 38 35
Screenshot 2025-10-08 at 10 38 11


However we are lazy loading that part after the profiler runs. 

At the same time I saw in this case and others that it appears that CORE functionality for LCP run along ours, and in this case they set the wrong LCP, is not a big deal, but I was wondering  if this might be the reason we don't properly set the bg image we identified as LCP (preload + fetch priority high ) 

2. the detection script detect 09439917-listing-lg.png image as loading with the correct size with if in the network tab we see the wrong size was requested. I think it's happening because the srcset is correct, but still the image is not loading correctly.
Screenshot 2025-10-08 at 10 46 34

selul:
@ineagu fixed the wpshouts things, fyi on the caching issue kinsta did had compatibility with wprocket on clear cache so they work in sync but only global and not per url as we need for profiler. 

will look into the. lcp ones.

ineagu:
> @ineagu fixed the wpshouts things, fyi on the caching issue kinsta did had compatibility with wprocket on clear cache so they work in sync but only global and not per url as we need for profiler, so I had to add dedicated compatibility with per url.
> 
> will look into the. lcp ones.

great, we don't need to cover more now. 

Found another use-case with our dear theme: 

https://lighter-puma-019eef.vertisite.cloud/

The LCP image is a background image in css, which Optimole is not replacing. However our profiler is correcly identifying it as LCP, shouldn't we be able to preload it in this case, even if is not a optimole url? Thoughts?

Not sure if replacing is also fixable, this is set using otter section as a background image, would be helpful if Otter is compatible with Optimole.

selul:
> However our profiler is correcly identifying it as LCP, shouldn't we be able to preload it in this case, even if is not a optimole url

No, it only considers Optimole URLs. Adding it will open up different risks and complexities which we need to account.

selul:
@ineagu from my testing is not the same, I was able to view this correctly on your page, the bg image is preloaded.

selul:
> With both enabled correct?

yes

PR #978, opened 2025-10-01
Improve quality setting handling in Compression settings 

ineagu:
Refactored the Compression.js component to better handle toggling between auto and manual quality settings while fixing https://github.com/Codeinwp/optimole-service/issues/1516#issuecomment-3355204589. Updated the minimum allowed quality value in settings.php from 1 to 50 to stay consistent with the front-end. 

Changed default compression setting to 80, which is more likely to be used vs 90, which is not recommended.

### All Submissions:

 Refactored the Compression.js component to better handle toggling between auto and manual quality settings while fixing https://github.com/Codeinwp/optimole-service/issues/1516#issuecomment-3355204589. Updated the minimum allowed quality value in settings.php from 1 to 50 to stay consistent with the front-end. 

Changed default compression setting to 80, which is more likely to be used vs 90, which is not recommended.

Closes [#](https://github.com/Codeinwp/optimole-service/issues/1516) .

1. Disable machine learning compression & save
2. Ensure the default value is used on the frontend.
3. Make sure going back to mauto or migrating from an older version works as expected.

### Other information:

* [x ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ x] Have you successfully ran tests with your changes locally?


inc/settings.php
```diff
@@ -278,7 +280,7 @@ public function parse_settings( $new_settings ) {
 					$sanitized_value = $this->to_bound_integer( $value, 100, 5000 );
 					break;
 				case 'quality':
-					$sanitized_value = $this->to_bound_integer( $value, 1, 100 );
+					$sanitized_value = $this->to_bound_integer( $value, 50, 100 );
```
Soare-Robert-Daniel, line 283: I think here is the most important thing, the initial issue with value being saved as one is because `quality` can be a string, `auto`, or number, to which the sanization does not consider the string case, thus the value will be saved as `1`, it is left untouched.

We should handle this case.

Soare-Robert-Daniel (changes requested):
I left some comments.

ineagu:
works as expected for me

PR #1094, opened 2026-07-21
Prevent duplicate scheduling of image processing actions

girishpanchal30:
### All Submissions:

Added a check to verify whether the `optml_start_processing_images` action is already scheduled. If it isn't, the action is scheduled. This prevents multiple instances of the same action from being scheduled.

Closes https://github.com/Codeinwp/optimole-wp/issues/1074

### Other information:

### DIFF
--- a/inc/media_offload.php
+++ b/inc/media_offload.php
@@ -1899,18 +1899,21 @@ class Optml_Media_Offload extends Optml_App_Replacer {
 					'action' => $type,
 				];
 			}
-			$total = ceil( $count / $batch );
-			self::schedule_action(
-				time(),
-				'optml_start_processing_images',
-				[
-					$action,
-					$batch,
-					1,
-					$total,
-					$step,
-				]
-			);
+
+			if ( ! self::is_scheduled( 'optml_start_processing_images' ) ) {
+				$total = ceil( $count / $batch );
+				self::schedule_action(
+					time(),
+					'optml_start_processing_images',
+					[
+						$action,
+						$batch,
+						1,
+						$total,
+						$step,
+					]
+				);
+			}
 		}
 
 		$response = [
--- a/inc/media_offload.php
+++ b/inc/media_offload.php
@@ -211,8 +211,9 @@ class Optml_Media_Offload extends Optml_App_Replacer {
 	 * @return void
 	 */
 	public function maybe_reschedule() {
+		$lock = get_transient( self::TRANSFER_LOCK_TRANSIENT );
 		// If this is in pending, we do nothing.
-		if ( self::is_transfer_lock_active() ) {
+		if ( false !== $lock ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


inc/media_offload.php
```diff
@@ -1970,6 +1988,165 @@ public static function is_scheduled( $hook ) {
...
+			);
+
+			if ( $updated ) {
+				wp_cache_delete( self::TRANSFER_LOCK_OPTION, 'options' );
+				return $token;
+			}
+		}
+
+		return false;
+	}
```
Soare-Robert-Daniel: Is there a more elegant way than this?

girishpanchal30: Yes, if the existing lock has expired, we delete it as a best-effort operation. It doesn't matter if 0 rows are affected, as another worker may have already cleared or replaced the lock. We then always attempt a plain `$wpdb->insert()`.

Since the outcome of the `INSERT` is the only thing that determines lock ownership, there is no need for a compare-and-swap (CAS)-conditioned `UPDATE` or a retry loop.

Please let me know if you have any other thoughts or suggestions.

Thanks!

Soare-Robert-Daniel: I see that WordPress has a similar mechanism when upgrading https://developer.wordpress.org/reference/classes/wp_upgrader/create_lock/

Do you think we can create something similar based on their pattern (you can tell the AI to work based on it)? For the moment, the `$wpdb->suppress_errors( $suppress );` does not inspire much safety.

And also, regarding the solution, we should add some tests to simulate this scenario.

girishpanchal30: I have implemented it as you suggested. Please check it and let me know if any changes are needed.

selul:
@girishpanchal30 have you found out the root cause? the scheduling already has enough locking check via wp-cron/action scheduler, i see we reinvent a lot of stuff.

girishpanchal30:
@selul I wasn't able to reproduce the issue in my local environment. However, after reviewing the code, I identified the underlying cause: when `number_of_images_and_pages()` is called during a [media offload](https://github.com/Codeinwp/optimole-wp/blob/master/inc/rest.php#L824-L836), it schedules the `optml_start_processing_images` action without ensuring that another processing chain hasn't already been scheduled for the same transfer.

As suggested by [Copilot](https://github.com/Codeinwp/optimole-wp/pull/1094#discussion_r3621829941), I implemented a locking mechanism that guards scheduling for both the `offload_images` and `rollback_images` actions. This ensures that only one processing chain can be scheduled per transfer, eliminating the race condition in which two concurrent requests could otherwise both schedule the same job before either completes.

selul:
@girishpanchal30 in this case we can use a simple transient which clears up when the process finish and is setting up when the process start.

girishpanchal30:
@selul I have implemented transient as you suggested with the latest commit. Please check it and let me know if any changes are needed.

PR #1133, opened 2026-09-01
fix: coerce object-shaped page profiles before viewport lookup

### LINKED ISSUE #1122: Viewport lazy-load profiling crashes on object-shaped profile data
## Summary

Viewport-based lazy-load processing can terminate a frontend request when stored page-profile data is object-shaped rather than array-shaped.

**Expected behavior:** Invalid or unavailable profile data is treated as unavailable, allowing image processing to continue without a fatal error.

**Actual behavior:** The page-profiler lookup accesses the object as an array and throws `Cannot use object of type stdClass as array`.

**Impact:** Affected frontend page requests can fail during Optimole image replacement.

## Customer context

- **Product / area:** Optimole WordPress plugin, page profiler and viewport lazy loading
- **Version:** 4.2.11
- **Environment:** WordPress 7.1, PHP 8.3.31
- **Integration / third party:** An external object cache may be in use; this is inferred from the plugin's storage selection and is not confirmed by telemetry.
- **Reported error / symptom:** `Cannot use object of type stdClass as array` at `inc/v2/PageProfiler/Profile.php:369`
- **Impact:** 10 telemetry occurrences across 2 production sites between 2026-08-27 and 2026-08-28 during frontend requests.

## Reproduction notes

1. Enable the viewport-based lazy-load workflow so `Optml_Lazyload_Replacer::can_lazyload_for()` consults page-profile data.
2. Arrange for the current profile's mobile or desktop storage value, or its `af` member, to be object-shaped rather than an array.
3. Process frontend HTML containing an eligible image.

**Observed in production:** the lookup throws `Cannot use object of type stdClass as array` from the reported Profile.php location. The exact storage backend and data origin are unavailable.

### Conclusion
Telemetry directly records the exception in the viewport-profile lookup. In `v4.2.11`, that lookup assumes each stored device profile and its `af` value are arrays, while the selected storage reader returns its value without runtime shape validation. An object-shaped value therefore reaches array access and produces the captured error. The internal REST write path constructs arrays, so the source of the object-shaped stored value remains unconfirmed.

### Where this likely occurs
- `inc/manager.php` — `Optml_Manager::replace_content()` lines 446-482 enables page profiling, loads current profile data, then processes frontend image content.
- `inc/lazyload_replacer.php` — `Optml_Lazyload_Replacer::can_lazyload_for()` lines 461-478 calls `Profile::is_in_all_viewports()` for viewport lazy loading before deciding whether an image is lazy-loaded.
- `inc/v2/PageProfiler/Profile.php` — `Profile::set_current_profile_data()` lines 336-347 copies each storage value into the current profile without type normalization; `Profile::is_in_all_viewports()` lines 362-375 then accesses `['af'][$image_id]`.
- `inc/v2/PageProfiler/Storage/ObjectCache.php` — `ObjectCache::get()` lines 54-62 returns `wp_cache_get()` directly despite the documented `array|false` return contract.
- `v4.2.11` is release commit `e9ee20c5a5e10565d2f8ffc56aa77fe76b202332`; the failing `is_in_all_viewports()` access is present there and unchanged through the inspected HEAD. The lookup originated in `d096d4bb28e01ddc56c5fce937adbd20d4b0fa63`.

### Engineering notes
- The profiler selects `Storage\ObjectCache` when WordPress reports an external object cache, or `Storage\Transients` otherwise, via `Profile::__construct()` in `inc/v2/PageProfiler/Profile.php` lines 89-106.
- `Profile::store()` writes the expected device data as arrays containing `af`, `bg`, and `lcp` in `inc/v2/PageProfiler/Profile.php` lines 183-218. `Optml_Rest::optimizations()` also normalizes its collected payload before calling that method at `inc/rest.php` lines 935-1046.
- The profiler storage class is filterable through `optml_page_profiler_storage`; a custom storage implementation is another possible source of an object-shaped value. Neither cache backend behavior nor a custom storage integration was available for inspection.
- The failure occurs before the existing `is_data_available()` result is used in `can_lazyload_for()`, so that later availability check cannot prevent this exception for malformed non-empty device values.

### Test coverage status
`tests/test-lazyload-viewport.php` covers normal array-backed profile lookup in `test_is_in_viewport_detection()` lines 179-203, stored structure in `test_profile_data_storage_and_retrieval()` lines 211-247, and missing or partial device data in lines 302-343. No relevant coverage was found during inspection for object-shaped cache values, malformed profile storage, or custom page-profiler storage. No test suite was run.

### What to verify or explore next
- Reproduce with a page-profile storage value whose device entry or `af` member is an object, under viewport lazy loading.
- Check affected sites' active object-cache drop-ins and whether the `optml_page_profiler_storage` filter is present.
- Run the page-profiler viewport test suite with array, false, and object-shaped stored values.
- Confirm whether transient-backed storage can encounter the same serialized data shape.

### Unknowns / follow-up
- Telemetry does not identify which stored profile field was object-shaped or how it entered storage.
- The affected sites' cache backend, installed extensions, and profiler configuration are unavailable.
- Git history establishes that the unsafe access predates 4.2.11 but does not establish a previously working release boundary.

## Confidence

**Confidence:** 94/100

Production telemetry records 10 frontend crashes on two sites in Optimole 4.2.11, and the reported source line directly indexes a stored profile value as an array without validating its documented `array|false` storage contract. The relevant call path is reachable during frontend viewport lazy-loading.

## Crash telemetry

| | |
|---|---|
| Occurrences | 10 |
| Distinct sites | 2 |
| First seen | 2026-08-27 10:11 UTC |
| Last seen | 2026-08-28 17:29 UTC |
| Crash location | `product:inc/v2/PageProfiler/Profile.php:369` |
| Request context | frontend |
| Inside Themeisle SDK | no |
| Product versions | 4.2.11 |
| WP versions | 7.1 |
| PHP versions | 8.3.31 |
| SDK versions | 3.3.58 |

---
**Source:** automated crash report — optimole-wp, fingerprint `3025c25bb349ed9cf9cdbed87b66cec3`
*Generated by bug-report-triage (ID: bug-report-triage_6a9275be3e9dc7.49048068)*

selul:
## Summary
- Viewport lazyload fatals when stored page-profile data is `stdClass` instead of an array (`Cannot use object of type stdClass as array` at `Profile.php:369`). That is the usual shape when an object-cache backend JSON-decodes without associative arrays.
- Normalize storage `get()` through `Storage\Base::normalize_value()`, coerce again when loading current profile data, and guard `af`/`lcp`/global lookups plus personalized background CSS so a bad shape never 500s the frontend.
- Valid object-shaped payloads are converted and still used (above-fold detection keeps working); corrupt scalars are treated as a cache miss.

Fixes #1122

## Test plan
- [x] `composer phpunit -- --filter='Test_Page_Profiler_Shape|Test_Lazyload_Viewport'` (42 tests)
- [x] PHPCS and PHPStan on the changed v2 files
- [ ] Confirm a Redis/Memcached drop-in that JSON-decodes objects no longer fatals frontend viewport lazyload
- [ ] With viewport lazyload on, a page with stored profile data still skips above-fold images after an object-cache roundtrip
- [ ] Missing/corrupt cache values still fall back to lazyloading all images

Made with [Cursor](https://cursor.com)

### DIFF
--- a/inc/v2/BgOptimizer/Lazyload.php
+++ b/inc/v2/BgOptimizer/Lazyload.php
@@ -3,7 +3,6 @@
 namespace OptimoleWP\BgOptimizer;
 
 use OptimoleWP\PageProfiler\Profile;
-use OptimoleWP\PageProfiler\Storage\Base as ProfilerStorage;
 use OptimoleWP\Preload\Links;
 use Optml_Lazyload_Replacer;
 
@@ -35,21 +34,8 @@ class Lazyload {
 		$css_selectors = [];
 		$preload_urls = [];
 		foreach ( Profile::get_active_devices() as $device ) {
-			$device_data = $data[ $device ] ?? [];
-			if ( ! is_array( $device_data ) ) {
-				$device_data = ProfilerStorage::normalize_value( $device_data );
-				$device_data = is_array( $device_data ) ? $device_data : [];
-			}
-			$personalized_selectors = $device_data['bg'] ?? [];
-			if ( ! is_array( $personalized_selectors ) ) {
-				$personalized_selectors = ProfilerStorage::normalize_value( $personalized_selectors );
-				$personalized_selectors = is_array( $personalized_selectors ) ? $personalized_selectors : [];
-			}
-			$lcp_data = $device_data['lcp'] ?? [];
-			if ( ! is_array( $lcp_data ) ) {
-				$lcp_data = ProfilerStorage::normalize_value( $lcp_data );
-				$lcp_data = is_array( $lcp_data ) ? $lcp_data : [];
-			}
+			$personalized_selectors = $data[ $device ]['bg'] ?? [];
+			$lcp_data = $data[ $device ]['lcp'] ?? [];
 			if ( OPTML_DEBUG ) {
 				do_action( 'optml_log', 'personalized_selectors: ' . $device . ' ' . print_r( $personalized_selectors, true ) );
 				do_action( 'optml_log', 'LCP data: ' . $device . ' ' . print_r( $lcp_data, true ) );
--- a/inc/v2/BgOptimizer/Lazyload.php
+++ b/inc/v2/BgOptimizer/Lazyload.php
@@ -3,6 +3,7 @@
 namespace OptimoleWP\BgOptimizer;
 
 use OptimoleWP\PageProfiler\Profile;
+use OptimoleWP\PageProfiler\Storage\Base as ProfilerStorage;
 use OptimoleWP\Preload\Links;
 use Optml_Lazyload_Replacer;
 
@@ -34,8 +35,21 @@ class Lazyload {
 		$css_selectors = [];
 		$preload_urls = [];
 		foreach ( Profile::get_active_devices() as $device ) {
-			$personalized_selectors = $data[ $device ]['bg'] ?? [];
-			$lcp_data = $data[ $device ]['lcp'] ?? [];
+			$device_data = $data[ $device ] ?? [];
+			if ( ! is_array( $device_data ) ) {
+				$device_data = ProfilerStorage::normalize_value( $device_data );
+				$device_data = is_array( $device_data ) ? $device_data : [];
+			}
+			$personalized_selectors = $device_data['bg'] ?? [];
+			if ( ! is_array( $personalized_selectors ) ) {
+				$personalized_selectors = ProfilerStorage::normalize_value( $personalized_selectors );
+				$personalized_selectors = is_array( $personalized_selectors ) ? $personalized_selectors : [];
+			}
+			$lcp_data = $device_data['lcp'] ?? [];
+			if ( ! is_array( $lcp_data ) ) {
+				$lcp_data = ProfilerStorage::normalize_value( $lcp_data );
+				$lcp_data = is_array( $lcp_data ) ? $lcp_data : [];
+			}
 			if ( OPTML_DEBUG ) {
 				do_action( 'optml_log', 'personalized_selectors: ' . $device . ' ' . print_r( $personalized_selectors, true ) );
 				do_action( 'optml_log', 'LCP data: ' . $device . ' ' . print_r( $lcp_data, true ) );
--- a/inc/v2/PageProfiler/Profile.php
+++ b/inc/v2/PageProfiler/Profile.php
@@ -226,9 +226,7 @@ class Profile {
 	 * @return array{w: int, h: int}|array{} The missing dimensions.
 	 */
 	public function get_missing_dimensions( int $image_id ): array {
-		$dimensions = self::get_device_member( self::DEVICE_TYPE_GLOBAL, 'm' )[ $image_id ] ?? [];
-
-		return is_array( $dimensions ) ? $dimensions : [];
+		return self::$current_profile_data[ self::DEVICE_TYPE_GLOBAL ]['m'][ $image_id ] ?? [];
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


selul:
Reviewed the fix against #1122, with the full suite run locally. The diagnosis and the fix are correct — but about 75% of the diff is provably redundant, and the PR can be trimmed to the storage boundary alone.

## What checks out

- The diagnosis matches the telemetry in #1122: the fatal is the `['af'][$image_id]` lookup in `is_in_all_viewports()`, reached when a JSON-decoding object-cache backend returns `stdClass`.
- `normalize_value()` is correctly written: recursive `get_object_vars()`, scalars treated as a cache miss, numeric string keys canonicalize to int on array insertion, so `af` lookups by int image ID keep working.
- Normalizing both backends is right — transients also ride the object cache when a drop-in is present.
- Full suite passes on the branch (275 tests locally), and the change quietly fixes a pre-existing unguarded `['lcp']['imageId']` access.

## The fix can be ~30 lines instead

`Profile::$current_profile_data` has exactly two writers: `set_current_profile_data()` (fed by `storage->get()`) and `reset_current_profile()` (`[]`). `exists()`, `get_profile_data()`, and `get_personalized_css()` (via `get_current_profile_data()`) all trace back to the same two storage `get()` methods. So the complete production fix is layer 1 only:

- `Base::normalize_value()`
- the one-line wrap in `ObjectCache::get()`
- the one-line wrap in `Transients::get()`
- plus the `?? null` guard in `is_lcp_image_in_all_viewports()` (genuine pre-existing bug, worth keeping)

**Verified empirically**: with only the `Storage/*` changes kept and `Profile.php` + `BgOptimizer/Lazyload.php` reverted to the development versions, every new test that drives data through a real entry point still passes — object cache, transients, `set_current_profile_data()`, `get_profile_data()`, `exists()`, and the end-to-end `replace_content()` with object-shaped transients. The only 8 failures are the tests that inject `stdClass` into the private static **via reflection** — a state production cannot reach once the boundary normalizes. No other suite regressions.

The extra layers also aren't free: `get_device_member()` → `get_device_data()` runs per image per device inside `can_lazyload_for()` — the hottest loop of viewport lazyload — replacing a direct `?? false` chain with two function calls plus `is_array` checks per lookup. The old lookup code is already safe against `false`/missing values (`empty()` guard + `??` isset semantics emit no warnings on `false['af']`).

Suggested trim: keep `Storage/*` + the `lcp` guard, drop the other `Profile.php`/`BgOptimizer` changes, and rewrite the 8 reflection-injection tests to seed through storage instead — they currently exercise states the fix itself makes impossible.

## Process flags

1. **Base branch is `master`** while the repo flow targets `development` (dependabot #1128, the sync commits). Fine if this is an intentional hotfix release; otherwise retarget.
2. **The `composer.lock` hunk** (themeisle-sdk 3.3.58 → 3.3.61) is unrelated to the fix and duplicates 405dee1 already on development. If retargeted it conflicts; if it stays a master hotfix, bundling a dependency bump into a crash fix adds release risk. Suggest dropping it either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

selul:
Addressed the review:

- Production fix is now the storage boundary (`normalize_value()` on `ObjectCache::get()` / `Transients::get()`) plus the `?? null` LCP `imageId` guard.
- Reverted the extra `Profile` / `BgOptimizer` lookup layers.
- Reflection-injection tests now seed object-shaped payloads through transients and load them via `set_current_profile_data()`.
- Retargeted the PR from `master` to `development`, so the unrelated themeisle-sdk bump is no longer in the diff.

`Test_Page_Profiler_Shape` + `Test_Lazyload_Viewport` pass (43 tests).

